PR #11, opened 2015-11-22
Adding firmware support for known wireless dongles

gkkpch:


volumio:
Great work! Why not putting firmwares in recipes?

gkkpch:
I wasn't sure what firmware comes with the raspberry kernel/platform files.
Debian seems to have most of the common firmware modules as packages.
Let's move it to recipes...

volumio:
I'll make a build now and see

PR #15, opened 2015-11-27
Update build.sh

learntofly83:
previous discussion about --recursive vs svn checkhout

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -100,7 +100,7 @@ mount /proc build/$BUILD/root/proc -t proc
 mount /sys build/$BUILD/root/sys -t sysfs
 echo 'Cloning Volumio'
 mkdir build/$BUILD/root/volumio
-git clone https://github.com/volumio/Volumio2.git build/$BUILD/root/volumio
+git clone --recursive https://github.com/volumio/Volumio2.git build/$BUILD/root/volumio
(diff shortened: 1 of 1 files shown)


volumio:
I cannot try it as of now... what does --recursive do?

learntofly83:
The www folder in the ui repo is set up in a funny way that you either needed to check it out with svn, recurse In to
It or there was some talk of submodules but it's a bit above my head... Either way the changes I've proposed produce a bigger squash FS file... But please test as I am still unable to produce a build that boots.. Also suggest using ext4 and going with the convention of the root living on /Dev/mmcblk0p2 not p3 as raspi-config won't resize non standard partition maps... Also tinkering win the 1.55 release, the fat32 partition is not flagged as 32LBA partition code in fdisk.. I think this places restrictions via the mbr on addressing past a certain amount of GB... Ie large cards might not work but again I'm not an expert

volumio:
As for the recursive, let's find a way to get Continuous Integration done first, then we can see what can work best... What do you think?

Then, for raspi-config, we don't need what you propose: 
- P1 is boot
- P2 is a filesystem which stores the squashfs images 
- P3 is the overlayfs partition which handles data, and its automatically expanded on first boot...

PR #16, opened 2015-11-28
Update raspberryconfig.sh

balbuze:
Section "adding raspi-config"
remove "/" before http://archive......


volumio:
Thank you very much man!

PR #40, opened 2016-02-13
Initial Odroid-C2 support

gkkpch:


volumio:
Great work gè, I'll start building now...
I may suggest to use lowercase naming for devices, as they are used also for the auto-publishing script I'm building now...
I know it may not look great..

PR #64, opened 2016-06-16
Update cuboxiimage.sh

christiansteier:
Check if /mnt/ and /mnt/volumio exist

### DIFF
--- a/scripts/cuboxiimage.sh
+++ b/scripts/cuboxiimage.sh
@@ -43,8 +43,19 @@ sync
 
  
 echo "Copying Volumio RootFs"
-sudo mkdir /mnt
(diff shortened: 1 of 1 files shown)


gkkpch:
Christian, 
this script is not finished yet, you won't get a Volumio 2 Image.
In case you could and would like to contribute I can help (from end of next week onwards), just send me a PM.
Basicly you now have an image with the Volumio root filesystem only.
You will need a kernel version  3.18 or later (or backport the overlayfs filesystem, I did that for the odroid on a kernel version 3.14). 
You can use the PI or Odroid scripts as an example for configuring a complete Volumio build.
Look at the <device>image.sh and the <device>config.sh

PR #66, opened 2016-06-26
Added the init script from the Master branch

gkkpch:
Need this for building an Odroid image from the dev branch, otherwise the C1/C2 won't restart properly.
This fix does not need to be merged into master as it is already there


volumio:
I'll switch to master in a couple of days... So we won't need that merged... What do you think?

gkkpch:
Fully agree, I put this in for me as a reminder. My C2 with RC2 wasn't rebooting, then I remembered I fixed that after the dev branch was started.
So, no need to merge....

PR #67, opened 2016-07-06
Syslinux for BIOS, Grub for UEIFI and OpenBox/Chromium for Volumio kiosk

gkkpch:
Finally a first properly working hybrid X86 version.
It uses Syslinux for booting Legacy BIOS and Grub 2 for EFI. 
It covers both 32bit and 64bit EFI bootloaders with the 32bit kernel.
Tested on VirtualBox 5.0.24, 32bit and 64bit host, with 32bit and 64bit UEFI, then Asus 64bit Notebook,  HP 32bit Notebook (old, never booted with grub), 64bit Gigabyte motherboard (own build) and an Intel NUC. 

The fsck problems have been fixed in x86.sh after unmounting the vfat boot partition.
The cause is creating a syslinux mbr without having a backup boot sector (not sure yet why this happens, may be an incompatibility between syslinux and gpt partitions). Would not have done any harm as we do not really need the backup boot sector, the fsck message just looked ugly at boot time

### DIFF
--- a/recipes/x86 (hybrid grub uefi).conf
+++ b/recipes/x86 (hybrid grub uefi).conf
@@ -12,7 +12,7 @@ addimportant=false
 omitrequired=false
 
 [Base]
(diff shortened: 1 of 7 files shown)


volumio:
Marvellous.... Testing now!

PR #75, opened 2016-07-17
GPIO pin permission fix

tomatpasser:
Root is currently needed to control GPIO pins. This fix solves it.
I have tested this on RPi 2 B. The gpio control files in /sys/class/gpio are correctly owned by group 'gpio' with the fix applied.

The fix is from [https://github.com/raspberrypi/linux/issues/1117#issuecomment-134227514](https://github.com/raspberrypi/linux/issues/1117#issuecomment-134227514). I was told it was already applied but can not find it anywhere on here.


volumio:
You're right! I forgot that... Thanks

PR #78, opened 2016-08-11
Fixed moving GTP partition backup tabled (initrd)

gkkpch:
You may want to wait with this as solaar74 reported problems on reboot after a successful first start.
Not sure yet if this is boot-related, it fails on mounting devices AFTER initrd

### DIFF
--- a/recipes/x86.conf
+++ b/recipes/x86.conf
@@ -31,7 +31,7 @@ keyring=debian-archive-keyring
 suite=jessie
 
 [Utils]
-packages=nano psmisc git wget ethtool curl ca-certificates ntp rpl avahi-daemon avahi-discover libnss-mdns zsync git-core binutils ca-certificates curl busybox parted udisks udisks-glue policykit-1 less usbutils
+packages=nano psmisc git wget ethtool curl ca-certificates ntp rpl avahi-daemon avahi-discover libnss-mdns zsync git-core binutils ca-certificates curl busybox parted udisks udisks-glue policykit-1 less usbutils gdisk
 source=http://ftp.nl.debian.org/debian
 keyring=debian-archive-keyring
 suite=jessie
--- a/scripts/initramfs/init-x86
+++ b/scripts/initramfs/init-x86
@@ -145,26 +145,12 @@ echo "Initrd initialisation finished..."
 
 #
 # Step 1: As the image likely has been dd'ed to a bigger disk, 
-#         the GPT backup sits in the wrong sectors. Move it 
-#         to the end of the disk before doing anything else.
+#         the GPT backup sits in the wrong sectors, fix it 
+#         before doing anything else.
 # =============================================================
 #
-# only run this at first boot
-
-mkdir boot
-mount -t vfat ${BOOTPART} /boot
-if [ -e "/boot/resize-volumio-datapart" ]; then
-  echo "Compensating for flashing starter image to a bigger disk"
-  BOOT_DEVICE=$(echo ${BOOTPART} | sed 's/[0-9]*//g') 
-  gdisk ${BOOT_DEVICE} >/dev/null <<EOF
-x
-e
-w
-y
-EOF
-fi
-umount /boot
-rm -r /boot
+BOOT_DEVICE=$(echo ${BOOTPART} | sed 's/[0-9]*//g') 
+parted ${BOOT_DEVICE} print Fix > /dev/null 2>&1
 
 #
 # Step 2: Check for recovery image, if not found create it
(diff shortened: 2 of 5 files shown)


volumio:
Gé, is this safe to merge?

gkkpch:
no, not yet. I'm still waiting for final feedback after additional changes. solaar74, FYC and balbuze are going to test...

volumio:
Let me know when you're done with that... Also, what if we prepare the installer in the meanwhile?

gkkpch:
What did you have in mind for the installer?
I could try the following: a whiptail based menu script to select the hard drive volumio should be copied to, then after issueing a warning, use 'dd' to copy. I somehow need to get this menu to run first (rc.local), but only after selecting a corresponding boot entry. After succesfull installation, the boot menu on the hard drive would be set to "automatic" (immediate boot). 
You might have a better idea though...

volumio:
This is a good idea! But we can also offer it in the UI, under system. By having this script done in bash and reporting progress on stdout we can use it both ways. What do you think?

gkkpch:
I would prefer the UI, makes it more user friendly. With a proper warning, that he/she is overwriting the selected disk without being recoverable. Probably have to put a note in it too, that a backup might be advisable.

Also think we should immediate boot, having menus does not add anything at all, unless you want to use it to trigger the install programm (initramfs could create a flag file depening on boot grub/ syslinux boot parameters)


On 23.08.2016, at 16:20, Volumio <notifications@github.com> wrote:

This is a good idea! But we can also offer it in the UI, under system. By having this script done in bash and reporting progress on stdout we can use it both ways. What do you think?

volumio:
I can do it in the UI without any problem... But may I rely on you for the script? Agree also that the menu adds nothing other than waiting time... 
You know what would be fantastic? An animated loading screen ...

gkkpch:
Good news from balbuze, my debug initrd gets past the "dead" point on his medion notebook. I think the issue has been resolved, when confirmed, i'll clean up for a PR.

On 23.08.2016, at 17:41, Volumio <notifications@github.com> wrote:

I can do it in the UI without any problem... But may I rely on you for the script? Agree also that the menu adds nothing other than waiting time... 
You know what would be fantastic? An animated loading screen ...

gkkpch:
ok, then i need to know what the script(s) should look like.
probably use lsblk to list the disks (and remove the boot disk from the list).
lsblk returns well readable device data in a script-friendly format.
Then we need dd in a way that it spits out copy progress, don't know the parameter by head. Have you got an example script somewhere from the spotify install I can have a look at?



On 23.08.2016, at 17:41, Volumio <notifications@github.com> wrote:

I can do it in the UI without any problem... But may I rely on you for the script? Agree also that the menu adds nothing other than waiting time... 
You know what would be fantastic? An animated loading screen ...

volumio:
I think we could call it the clone\install function, and it may be well done for all devices. With this you can clone a volumio install (for backup purposes, why not...)  on a new drive.
It can be in phases, like a wizard: 
select the drive 
clone\install 
apply.

The difference beetween clone and install will be that with clone you'll just do a straight dd, with install you do a dd then delete user data and resize data part. 
What do you think?

gkkpch:
Sounds ok, let's do this....

gkkpch:
But remember, feature freeze :)

volumio:
Lol, yes you're right... Let's fix stuff and then implement it.... Let's see in a "parallel" way

gkkpch:
ok, can you merge this one? Removing the bootmenu for grub/uefi will be done with the next fix, for syslinux it was never there I noticed.

volumio:
Gè, guess we have a regression here... Cannot boot it anymore on any of my devices... Is there any additional dependency of the boot script? I'll try rebuilding it...

volumio:
Nope... seem not to work... If you want to try it:
http://updates.volumio.org/x86/volumio/0.986/volumio-0.986-2016-08-26-x86.img.zip

gkkpch:
There are new dependencies,  gdisk was added to the recipes for using it in init-86.
==>> libstdc++.so.6 was added too, perhaps you need that one on the host??

gkkpch:
it is in recipes, so that can't be it.

gkkpch:
checked my build from" volumio/Build". It boots legacy and uefi (but volumio.service is not starting)
flashed your build from above. It boots legacy and uefi, same issue with volumio.service.
works on my Gigabyte, HP Notebook, Asus Notebook but..... not on Intel NUC, bugger!!!!

gkkpch:
is it possible to revert?

gkkpch:
just wait, starting with a version from 15.07 to see where it went wrong

gkkpch:
No need to revert!!!!!!!!
Nothing wrong with the build, the problem is another one :smile: :smile: :smile: :smile:
You're going to like this one, it is booting fine from the new image, but not the way we expected:
  - we both had a volumio install on the nuc's or notebook's hdd, right?
  - now, when we boot, how do we locate our disks devices? Get it??

gkkpch:
changing by-label back to by-uuid and fixing fstab at boot time

PR #79, opened 2016-08-12
cleanups in build.sh

xipmix:
I have been experimenting with building and noticed a few small things that could be improved.


volumio:
Thanks, well done

PR #80, opened 2016-08-15
Make calling rpi-update more robust

xipmix:
I noticed that raspberryconfig.sh saves 'rpi-update' to a specific location but then calls it without that full path. Also, the code tries to self-update and self-execute, which seems like a complication we could do without. As I've been experimenting with the build code I had some runs fail to run rpi-update,
so I put this PR together to address the issue.


xipmix:
Please ignore this request. When I tested the change more carefully I found the actual problem - curl has become fussier about SSL certificates, so the download doesn't actually occur.

PR #82, opened 2016-08-19
Make sure the dtoverlay binaries have somewhere to go

xipmix:
At this point in the process, /mnt/volumio/rootfs/opt (copied from build/arm/root/opt) is empty.
Create the destination directory before copying the dtoverlay programs into it.

Without this change I get this error during build:
```
echo "Custom dtoverlay pre and post"
cp: target ‘/mnt/volumio/rootfs/opt/vc/bin/’ is not a directory
```


volumio:
Thanks!

PR #84, opened 2016-08-28
x86 boot improvement & initial cubox-i support

gkkpch:
x86: this will allow you to boot a different build from usb after you cloned the original to the pc's hdd. However, after you cloned and you try to boot the same build from usb, it will always take the pc's hdd. Which is ok in most cases. The only real way of solving it, would be to make either the uuid or label unique on first boot or after an ota update and update syslinux config, grub config and fstab. The ota update would need a flag file, as would a new build.
For the time being, it will work.


volumio:
Tested and works as you suggested on x86...

PR #90, opened 2016-09-24
Generate SSH host keys on first start

Wraul:
The host keys should not be reused between multiple devices.
This change ensures that each Volumio device have unique host keys.

If you know of a better way to do this please let me know and I change the pull request.

I have tested the change on a Raspberry Pi 3.

### DIFF
--- a/scripts/volumioconfig.sh
+++ b/scripts/volumioconfig.sh
@@ -334,6 +334,9 @@ echo "Setting Mpd to SystemD instead of Init"
 update-rc.d mpd remove
 systemctl enable mpd.service
 
+echo "Removing default SSH host keys"
+# These should be created on first boot to ensure they are unique on each system
+rm -v /etc/ssh/ssh_host_*
 
 #####################
 #Audio Optimizations#-----------------------------------------
--- a/volumio/bin/firststart.sh
+++ b/volumio/bin/firststart.sh
@@ -5,6 +5,9 @@ echo "Volumio first start configuration script"
 echo "configuring unconfigured packages"
 dpkg --configure --pending
 
+echo "Generating SSH host keys"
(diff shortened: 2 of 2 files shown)


volumio:
Yes, this should work perfectly. I always wondered why this would be necessary... Could you point me in that direction?

Wraul:
The host key is mainly used to help the client identify the server and to verify that it is the correct one. This protects against man in the middle attacks.

When you use the same host key on all Volumio installations and include it in the image, the key becomes public knowledge. This allows an adversary to take this publicly known key and impersonate a Volumio device. By combining this with other attacks the adversary might trick a user into believing they are connecting to their Volumio device when they in reality is connecting to a server controlled by the adversary.

After this change all Volumio devices will have their own key which is known only to them. That way it is impossible to take a key from one Volumio device to impersonate another.

All of this is kind of theoretical though in the case of Volumio.
The fact that the same password is used for all Volumio installations undermines most of the security improvement gained by using unique host keys.

I hope this at least makes things slightly clearer.

volumio:
Yes, thank you. Your explanation was really clear.
Security is a concern for us, would you like to help us fix other flaws? Of course we must find the right balance beetween security and usability...

Wraul:
Regarding your inquiry about me working on fixing more security flaws.

Unfortunately I don't have that much time to work on a project like this.
But I can at least list a couple of changes that I believe would improve security.
- Remove the default root password.
- Disable SSH by default and let the user enable it in the GUI. Preferably it should be possible to configure keybased authentication.
- Protect the settings pages in the GUI with a password. Preferably a password that the user is forced to set before it's possible to access any parts of the GUI. This could be part of a nice first run wizard.

PR #97, opened 2016-10-27
Install a minimal set of locales

xipmix:
It's useful to have some locales, but the full set can cost as much as 60-100MB
of disk space. Use localepurge to strip this back to a minimal set (enough to
keep Perl happy) at a cost of around 3MB.

Tested using ```-b arm -d pi ``` and ```-b x86 -d sparky``` . I did not test-boot the final images, that would probably worth doing before merging this.

Closes: volumio/Volumio2#306

### DIFF
--- a/recipes/arm.conf
+++ b/recipes/arm.conf
@@ -24,7 +24,7 @@ keyring=debian-archive-keyring
 suite=jessie
 
 [Utils]
-packages=nano psmisc ethtool curl ca-certificates ntp git-core parted wget avahi-daemon avahi-discover libnss-mdns zsync udisks udisks-glue policykit-1 less fake-hwclock usbutils
+packages=nano psmisc ethtool curl ca-certificates ntp git-core parted wget avahi-daemon avahi-discover libnss-mdns zsync udisks udisks-glue policykit-1 less fake-hwclock usbutils locales localepurge
 source=http://archive.raspbian.org/raspbian
 keyring=debian-archive-keyring
 suite=jessie
--- a/recipes/arm64.conf
+++ b/recipes/arm64.conf
@@ -24,7 +24,7 @@ keyring=debian-archive-keyring
 suite=jessie
 
 [Utils]
-packages=nano psmisc ethtool curl ca-certificates ntp git-core parted wget avahi-daemon avahi-discover libnss-mdns zsync udisks udisks-glue policykit-1
+packages=nano psmisc ethtool curl ca-certificates ntp git-core parted wget avahi-daemon avahi-discover libnss-mdns zsync udisks udisks-glue policykit-1 locales localepurge
 source=http://ftp.ch.debian.org/debian
 keyring=debian-archive-keyring
 suite=jessie
--- a/recipes/x86.conf
+++ b/recipes/x86.conf
@@ -31,7 +31,7 @@ keyring=debian-archive-keyring
 suite=jessie
 
 [Utils]
-packages=nano psmisc git wget ethtool curl ca-certificates ntp rpl avahi-daemon avahi-discover libnss-mdns zsync git-core binutils ca-certificates curl busybox parted udisks udisks-glue policykit-1 less usbutils gdisk
+packages=nano psmisc git wget ethtool curl ca-certificates ntp rpl avahi-daemon avahi-discover libnss-mdns zsync git-core binutils ca-certificates curl busybox parted udisks udisks-glue policykit-1 less usbutils gdisk locales localepurge
 source=http://httpredir.debian.org/debian
 keyring=debian-archive-keyring
 suite=jessie
--- a/scripts/volumioconfig.sh
+++ b/scripts/volumioconfig.sh
@@ -7,6 +7,27 @@ export LC_ALL=C LANGUAGE=C LANG=C
 /var/lib/dpkg/info/dash.preinst install
 dpkg --configure -a
 
+# Reduce locales to just one beyond C.UTF-8
+echo "Existing locales:"
+locale -a
+echo "Generating required locales:"
+[ -f /etc/locale.gen ] || touch -m /etc/locale.gen
+echo "en_US.UTF-8 UTF-8" >> /etc/locale.gen
+locale-gen
+echo "Removing unused locales"
+echo "en_US.UTF-8" >> /etc/locale.nopurge
+# To remove existing locale data we must turn off the dpkg hook
+sed -i -e 's/^USE_DPKG/#USE_DPKG/' /etc/locale.nopurge
+# Ensure that the package knows it has been configured
+sed -i -e 's/^NEEDSCONFIGFIRST/#NEEDSCONFIGFIRST/' /etc/locale.nopurge
+dpkg-reconfigure localepurge -f noninteractive
+localepurge
+# Turn dpkg feature back on, it will handle further locale-cleaning
+sed -i -e 's/^#USE_DPKG/USE_DPKG/' /etc/locale.nopurge
+dpkg-reconfigure localepurge -f noninteractive
+echo "Final locale list"
+locale -a
+echo ""
 
 #Adding Main user Volumio
 echo "Adding Volumio User"


volumio:
Very good, like the way you handled it. Doing a test build now!

volumio:
It seems good, and the image is even smaller. However I get an issue with network shares discovery. Don't know if that's related, but seems the case...
Investigating..

volumio:
Not an issue, samba is not affected...
I must say: good work! We also saved 30MB of build size! well done!

macmpi:
there seem to still be some locale issues on 2.118: any hint?
```
volumio@volumio:~$ sudo apt-get install keyboard-configuration
Reading package lists... Done
Building dependency tree
Reading state information... Done
The following NEW packages will be installed:
  keyboard-configuration
0 upgraded, 1 newly installed, 0 to remove and 24 not upgraded.
Need to get 766 kB of archives.
After this operation, 2536 kB of additional disk space will be used.
Get:1 http://archive.raspbian.org/raspbian/ jessie/main keyboard-configuration all 1.123 [766 kB]
Fetched 766 kB in 1s (423 kB/s)
perl: warning: Setting locale failed.
perl: warning: Please check that your locale settings:
	LANGUAGE = (unset),
	LC_ALL = (unset),
	LC_CTYPE = "UTF-8",
	LANG = (unset)
    are supported and installed on your system.
perl: warning: Falling back to the standard locale ("C").
locale: Cannot set LC_CTYPE to default locale: No such file or directory
locale: Cannot set LC_ALL to default locale: No such file or directory
debconf: delaying package configuration, since apt-utils is not installed
Selecting previously unselected package keyboard-configuration.
(Reading database ... 18792 files and directories currently installed.)
Preparing to unpack .../keyboard-configuration_1.123_all.deb ...
Unpacking keyboard-configuration (1.123) ...
Processing triggers for systemd (215-17+deb8u6) ...
Setting up keyboard-configuration (1.123) ...
locale: Cannot set LC_CTYPE to default locale: No such file or directory
locale: Cannot set LC_ALL to default locale: No such file or directory
locale: Cannot set LC_CTYPE to default locale: No such file or directory
locale: Cannot set LC_ALL to default locale: No such file or directory
locale: Cannot set LC_CTYPE to default locale: No such file or directory
locale: Cannot set LC_ALL to default locale: No such file or directory
/usr/bin/locale: Cannot set LC_CTYPE to default locale: No such file or directory
/usr/bin/locale: Cannot set LC_ALL to default locale: No such file or directory
update-rc.d: warning: start and stop actions are no longer supported; falling back to defaults
update-rc.d: warning: start and stop actions are no longer supported; falling back to defaults
Processing triggers for systemd (215-17+deb8u6) ...
```

xipmix:
What does 'locale' report? I get something rather odd that I'd rather not share until I'm sure it helps the cause.

xipmix:
Initial guess: you are logging in over SSH and some of the environment is coming with you:
```
$ grep LC_ /etc/ssh/sshd_config 
AcceptEnv LANG LC_*
```
It seems one can override those settings like so:
```
volumio@volumio:~$ sudo update-locale LANG=en_US.UTF-8 LC_ALL=en_US.UTF-8 LANGUAGE=en_US:en
```
which updates /etc/default/locale, which I guess has an influence on all processes started thereafter. But that may not be the right way to approach this; after all ssh allows those variables to be passed in for a reason.

Firstly, I don't think volumio is trying to support the command-line use case. If the UI works, that's enough. And it seems level things _are_ working, just with a bit of locale noise.

That said, it may still be useful to allow the user to generate locale files for particular locales they want.

macmpi:
Yes I'm under SSH.
Here are the requested info below. Note that I did this both under SHH and directly on console attaching a USB keyboard. Command output are the same in both cases.
```
volumio@volumio:~$ locale
locale: Cannot set LC_CTYPE to default locale: No such file or directory
locale: Cannot set LC_ALL to default locale: No such file or directory
LANG=
LANGUAGE=
LC_CTYPE=UTF-8
LC_NUMERIC="POSIX"
LC_TIME="POSIX"
LC_COLLATE="POSIX"
LC_MONETARY="POSIX"
LC_MESSAGES="POSIX"
LC_PAPER="POSIX"
LC_NAME="POSIX"
LC_ADDRESS="POSIX"
LC_TELEPHONE="POSIX"
LC_MEASUREMENT="POSIX"
LC_IDENTIFICATION="POSIX"
LC_ALL=
volumio@volumio:~$ locale -a
locale: Cannot set LC_CTYPE to default locale: No such file or directory
C
C.UTF-8
POSIX
en_US.utf8
volumio@volumio:~$ cat /etc/default/locale
#  File generated by update-locale
```

xipmix:
I don't understand why LC_CTYPE is set that way. Mine is now en_US.UTF-8
You could set LC_ALL and possibly LANG in /etc/default/locale, try logging out and logging in.
It might be worth logging in on the console and finding what 'locale' reports in that environment.

macmpi:
As mentioned same result in console.
This is from a blank 2.118 image: something may need to be set in Build then.
You may have set it differently at some point after flashing it.

xipmix:
Yes, the LC_CTYPE setting I have is because I modified /etc/default/locale (with update-locale):
```
#  File generated by update-locale
LANG=en_US:UTF-8
LC_ALL=en_US.UTF-8
```
But I wonder at how yours got set to 'UTF-8'. That is just a character encoding. As far as I understand there should be a locale prefixed to it, ie ```ll_cc`` where ll is a language code and cc is a country code. LC_CTYPE is one of those variables that allow overriding of the LANG setting for a subset of localisation concerns.

Reference: https://www.gnu.org/software/gettext/manual/html_node/Locale-Environment-Variables.html#Locale-Environment-Variables

macmpi:
Not "mine": it's inherited straight from Built image (2.118 for instance), as you can easily check it.

xipmix:
Ah, thanks for the cluebat. However it got that way, I think LC_CTYPE is set incorrectly. I think this demonstrates that:
```
$ env |grep -c LC_
0
$ export LC_CTYPE="UTF-8"
bash: warning: setlocale: LC_CTYPE: cannot change locale (UTF-8): No such file or directory
$ env | grep LC
LC_CTYPE=UTF-8
```
So I think you've found a bug, which is good.

I think all this could be avoided if volumio to had a default locale setting, but I'm not sure if it has one set at present. I was expecting it to be en_US.UTF-8 but that seems not to be the case.

I think we can set it explicltly with /etc/default/locale, the questions are how and to what.

On the how, the gettext manual offers some clues
 * set LANG in /etc/default/locale.
   All the LC_* settings would follow this if they aren't set in the environment already.
   But however LC_CTYPE is getting set, it is overriding LANG.
 * alternative: set LC_ALL in /etc/default/locale
   Again all the LC_* settings would follow this, but they can't be overridden by user settings.

I think the right thing is to set LANG and fix whatever is messing up LC_CTYPE, but given that, what do we set it to?
 * Build/scripts/volumioconfig.sh sets 'C' at the start but does not try to enforce it within the build image.  We probably want a UTF-8 locale, anyway
 * en_US.UTF-8 seems the obvious choice since we build it explicitly in the volumioconfig.sh script.
 * can a case be made for POSIX?

xipmix:
Now here's an odd thing. I can't reproduce your result. I just downloaded and flashed 2.118 (for rpi). The checksum of the unzipped image is 647ac3ab955a5967135cdba7475d0030.
Fired it up and logged in:
```
% env | grep -c LANG
0
% env |grep -c LC_
0
% ssh volumio@volumio
volumio@volumio:~$ locale
LANG=
LANGUAGE=
LC_CTYPE="POSIX"
LC_NUMERIC="POSIX"
LC_TIME="POSIX"
LC_COLLATE="POSIX"
LC_MONETARY="POSIX"
LC_MESSAGES="POSIX"
LC_PAPER="POSIX"
LC_NAME="POSIX"
LC_ADDRESS="POSIX"
LC_TELEPHONE="POSIX"
LC_MEASUREMENT="POSIX"
LC_IDENTIFICATION="POSIX"
LC_ALL=
```
Exact same result from ```locale``` when I login to the console.

It would appear we are defaulting to 'POSIX', so perhaps we had the right answer there all along.

Can you try ```grep -rl LC_CTYPE /etc``` on your image to see if you can find where it might be getting set?

macmpi:
```
volumio@volumio:~$ sudo grep -rl LC_CTYPE /etc
volumio@volumio:~$
```

macmpi:
FYI: Build log on related matter:
```
Existing locales:
C
C.UTF-8
POSIX
Generating required locales:
Generating locales (this might take a while)...
  en_US.UTF-8... done
Generation complete.
Removing unused locales
localepurge: Disk space freed in /usr/share/locale: 69368 KiB
localepurge: Disk space freed in /usr/share/man: 3248 KiB

Total disk space freed by localepurge: 72616 KiB

Final locale list
C
C.UTF-8
POSIX
en_US.utf8
```

Issue #93, opened 2016-10-20
/var/log/samba missing because /var/log is overlaid with a tmpfs


xipmix:
I noticed a lot of complaints from samba-related daemons that /var/log/samba is missing. firststart.service creates it, and it exists in the fully-booted system on the first boot.

But on subsequent boots the /var/log entry in /etc/fstab 

```
tmpfs   /var/log                tmpfs   nodev,uid=1000,mode=0777,gid=4, 0 0
```

overlays what is in the distributed image with an empty tmpfs filesystem.

I can umount the tmpfs and underneath is all the expected /var/log structure. So I wonder if what is meant here is:

```
tmpfs   /var/log/volumio                tmpfs   nodev,uid=1000,mode=0777,gid=4, 0 0
```

i.e. reserve a special directory, owned by the volumio user, for the volumio app's log. Changing this would also require fixing the path used in /app/index.js.

Otherwise, perhaps  /var/log/samba should be created on every boot, perhaps as another tmpfs

```
tmpfs   /var/log/samba         tmpfs   defaults,nodev,noatime,mode=0755 0 0
```

Attached is a zip showing the first and second boots of a fresh 2.001, with the contents of /var/log once the system is fully started up.
[firsttwoboots.zip](https://github.com/volumio/Build/files/541441/firsttwoboots.zip)

macmpi:
Yes have similar issue here opening ssh sessions:
`Oct 27 14:34:13 volumio sshd[1379]: pam_env(sshd:session): Unable to open env file: /etc/default/locale: No such file or directory
Oct 27 14:34:13 volumio sshd[1379]: lastlog_openseek: Couldn't stat /var/log/lastlog: No such file or directory
Oct 27 14:34:14 volumio sshd[1379]: lastlog_openseek: Couldn't stat /var/log/lastlog: No such file or directory
`

volumio:
The intention is write all logs to ram, in order to prevent sd card corruption. 
At this poin I think we should add to fstab every log entry for complaining services... 
What do you think?

xipmix:
Thanks for clarifying. In that case it makes sense to add more fstab entries for different log dirs. A brief comment fstab about why they are there might be useful. However if some logs are going to be written by root and some by other users, I think it is worth keeping the non-root ones in separate /var/log subdirectories that can be set up with the right owner and permissions.

volumio:
Maybe we could create a service that creates those temp dirs with appropriate permissions... 
[...]

PR #99, opened 2016-11-01
Create subdirectories of /var/log

xipmix:
Issue #93 reported that /var/log was being created as a tmpfs on top of the SD card and that it lacked some subdirectories that the build system had placed there, e.g. /var/log/samba.
This series of changes ensures that the subdirs created during image build are reproduced on the tmpfs version of /var/log, once that filesystem is mounted at runtime.

This has been tested but as pieces, rather than making a full image and booting from it. Some independent build tests seem advisable.

Merging this will require moving /var/log/volumio.log to /var/log/volumio/volumio.log.


gkkpch:
HI,

@xipmix can you verify the following:
I started on pine64 with its own set of image and config file and would like to stay compatible with the rest. 
I only need to replace 
      tmpfs   /var/log                tmpfs   size=20M,nodev,uid=1000,mode=0777,gid=4, 0 0
with
      tmpfs   /var/log                tmpfs   size=20M,nodev 0 0
in fstab, correct?

I on my side need to look at the x86 implementation.
There we have an fstab template dependency, which needs to be checked.

xipmix:
On Tue, Nov 01, 2016 at 02:54:54PM -0700, Gé Koerkamp wrote:
> HI,
> 
> I only need to replace 
>       tmpfs   /var/log                tmpfs   size=20M,nodev,uid=1000,mode=0777,gid=4, 0 0
> with
>       tmpfs   /var/log                tmpfs   size=20M,nodev 0 0
> in fstab, correct?

Yes, I think that would be the only change to a pine64config.sh if
you are writing one.

> I on my side need to look at the x86 implementation.
> There we have an fstab template dependency, which needs to be checked.

Yes, I guess I forgot about that.

I also notice I missed some changes to scripts/configure.sh.

@volumio, please don't pull this yet.

> 
> -- 
> You are receiving this because you were mentioned.
> Reply to this email directly or view it on GitHub:
> https://github.com/volumio/Build/pull/99#issuecomment-257710564

--

xipmix:
I've now tested this more carefully - built a full rpi image and booted it - it works as expected, with the two addtional commits. Testing on x86 would still be a good idea.

volumio:
I wll test it as well tomorrow! Great work so far!

xipmix:
> I on my side need to look at the x86 implementation. There we have an fstab template dependency, which needs to be checked.

@gkkpch, I have looked into scripts/x86* and scripts/{volumio,}config.sh and the only fstab template dependency I can see is fixing up the uuid of the / partition. Am I missing something?

gkkpch:
Yes, there us a bit in the initrd, i'll see to it tonight

----- Reply message -----

gkkpch:
I believe it is OK

----- Reply message -----

gkkpch:
Tested x86 just to make sure, works fine.
@xipmix: job well done!

xipmix:
@volumio, I think this is ready now

volumio:
Sorry for taking long.. Still testing it... I will merge it asap!

xipmix:
> Sorry for taking long.. Still testing it... I will merge it asap! 

no worries, it's one you will want to be right.

macmpi:
will this fix this issue too? https://github.com/volumio/Volumio2/issues/960

macmpi:
Would be great to have it in.
One question though: with all logs mapped to RAM, how can we check/review events happening during shutdown/restart, as they will be gone after reboot?
Maybe we can add a setting in` /dev` page to store shutdown journalctl somewhere in flash for debug? Or any better idea?

volumio:
Adding a setting to dev page to dump logs is a great idea. I'll do that

xipmix:
@volumio I tried to fix the conflicts that have grown up over the last 3 months but now all of the changes I want to go in are being treated as conflicts. What should I do to fix this - make a new PR?

macmpi:
@volumio actually my point was somehow different, sorry for being unclear: not dumping but keeping journal across reboot in particular, so that one can check/monitor journal across reboots (in particular the shutdown part).
`/dev` setting could flip between current default [`/etc/systemd/journald.conf`](https://wiki.archlinux.org/index.php/Systemd#Journal) setting, or modify with `Storage=persistent` and apply `systemctl restart systemd-journald`.

@xipmix you may have to resync your branch from master, and submit again I'm afraid...

xipmix:
This PR is defunct, see #136 instead

PR #108, opened 2016-11-30
Initial Banana Pi M2U support (WIP)

gkkpch:


volumio:
It does look as a progress! Did you manage to boot with it?

PR #110, opened 2016-12-03
Remove a stray / that causes umount to fail. Fixes https://github.com/volumio/Volumio2/issues/778.

xipmix:


volumio:
Thanks!!! That one was hard to notice! You made my day sir!

Issue #102, opened 2016-11-11
Raspbian sources incomplete (and no mirror)


macmpi:
Hi,
While I was willing to install `python-rpi.gpio` (to implement button shutdown), I noted `apt-get install` would not make it available.
Looking at `/etc/apt/sources.list`, it seems we'd rather use std definition, as in std Jessie images:
```
deb http://mirrordirector.raspbian.org/raspbian/ jessie main contrib non-free rpi
deb-src http://archive.raspbian.org/raspbian/ jessie main contrib non-free rpi
```
Shall we fix it [in here](https://github.com/volumio/Build/blob/master/volumio/etc/apt/sources.list)?
This would also allow use of proper mirroring according to users' regions.
Thanks.

volumio:
Yes, we should. Thanks for reporting

macmpi:
You're welcome.
Unsure the deb-src is really necessary in normal use though, may comment/remove it.

volumio:
It's good to have, if people want to build stuff

macmpi:
thing is, even with that change (followed by `sudo apt-get update`), I can not install `python-rpi.gpio`
```
volumio@volumio:~$ sudo apt-get install python-rpi.gpio
Reading package lists... Done
Building dependency tree
Reading state information... Done
E: Unable to locate package python-rpi.gpio
E: Couldn't find any package by regex 'python-rpi.gpio'
```
Package is present in `http://archive.raspberrypi.org/debian/ jessie/main armhf Packages`, and is found on Jessie lite using same `/etc/apt/sources.list` ...weird.

macmpi:
Ok, so, in addition to updating `/etc/apt/sources.list`, one need to add the following `/etc/apt/sources.list.d/raspi.list`, containing the following:
```
deb http://archive.raspberrypi.org/debian/ jessie main ui
# Uncomment line below then 'apt-get update' to enable 'apt-get source'
#deb-src http://archive.raspberrypi.org/debian/ jessie main ui
```
(unsure what the existing empty `/etc/apt/sources.list.d/multistrap-debian.list` is meant for.

In addition, **key** for archive.raspberrypi.org/debian/ jessie should be added to keyring... (unsure this [hint](https://github.com/debian-pi/raspbian-ua-netinst/issues/64) may help for that).

With that, additional packages (like `python-rpi.gpio` are found and install fine).

macmpi:
[...]

PR #116, opened 2016-12-09
Complete Raspi sources

macmpi:
as per https://github.com/volumio/Build/issues/102

Please check sources.list.d/raspi.list  necessary key

### DIFF
--- a/scripts/raspberryconfig.sh
+++ b/scripts/raspberryconfig.sh
@@ -38,6 +38,14 @@ options snd-usb-audio nrpacks=1
 options snd-usb-audio index=5
 options snd_bcm2835 index=0" >> /etc/modprobe.d/alsa-base.conf
 
+echo "Adding Raspberrypi.org Repo"
+echo "
+deb http://archive.raspberrypi.org/debian/ jessie main ui
+deb-src http://archive.raspberrypi.org/debian/ jessie main ui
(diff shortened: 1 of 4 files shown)


volumio:
Do you have a hint?

macmpi:
does [this](https://github.com/debian-pi/raspbian-ua-netinst/issues/64) help for key-related stuff?

macmpi:
Probably this for the key thing:
```
wget http://archive.raspberrypi.org/debian/raspberrypi.gpg.key
apt-key add raspberrypi.gpg.key
```
Unsure when it shall end-up in  within the build scripts.

macmpi:
Now should be pretty much it.

volumio:
I think the raspberry pi repo should be copied over just to raspberry pi, in this commit
c2c3cda95f21e008a41a10773e9fe21a3f7ae535

macmpi:
Not sure I understand: you mean put all sources within single file `volumio/etc/apt/sources.list`, and cancel `volumio/etc/apt/sources.list.d/raspi.list` file?
(just did it like that to mimic what's in Jessie)

volumio:
No, that we should put volumio/etc/apt/sources.list.d/raspi.list only on raspberry pi device, so in raspberryconfig.sh

macmpi:
Ok, done.

volumio:
Perfect!

PR #117, opened 2016-12-10
Fix priviledges on /etc/hostapd/hostapd.conf

macmpi:
was probably a typo with incomplete file name


volumio:
Not a typo: we're setting permissions for the entire folder, not the file itself

PR #118, opened 2016-12-23
Links to Odroid kernels, see comment

gkkpch:
The README.md is not correct yet, but now holds the links to the used kernel sources for the different boards. 
You would need to clone/ download these and add them to the volumio repo and then replace the links in README.md

I have added the kernel source info to all README.md's of the various platform repos


gkkpch:
These are the pending commits, they have all the changes/ recipes to build with Debian armv7

PR #120, opened 2017-01-13
Use Google DNS as fallback only

macmpi:
Current setup sets Google's DNS as defaults.
This can raise privacy concerns for some users (yes Google does track usage), and prevents addressing local resources (NAS, others) with local DNS.
Proposal is to let resolv.conf use DHCP server provided DNS, and append Google ones at the end, in case initial ones fail.


volumio:
very well done!

macmpi:
Notably [fixes](https://github.com/volumio/Volumio2/issues/791) this too. (was a bit expected but could not test)

PR #126, opened 2017-01-15
[wireless] Use syslog for wpa_supplicant since stdout is closed

earlchew:
Signed-off-by: Earl Chew <earl_chew@yahoo.com>


volumio:
Sensible!

PR #129, opened 2017-01-15
[dynswap.sh] Only active swap file if there are no other swap areas

earlchew:
Signed-off-by: Earl Chew <earl_chew@yahoo.com>

### DIFF
--- a/volumio/bin/dynswap.sh
+++ b/volumio/bin/dynswap.sh
@@ -1,8 +1,9 @@
 #!/bin/sh
 
 RAMSIZE=`cat /proc/meminfo | grep MemTotal | awk '{ print $2 }'`
+SWAPDEVS=`cat /proc/swaps | wc -l`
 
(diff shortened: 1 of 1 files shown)


volumio:
What is the effect of this change? For situations different than yours...

earlchew:
@volumio - For the default configuration, there is no change in the current behaviour because when the system boots, `/proc/swaps` will be empty. It will only contain a heading:
```
$ cat /proc/swaps
Filename				Type		Size	Used	Priority
```
This means that condition `[${SWAPDEVS} -le 1 ]` will be true, and the `dynswap.sh` script will create the swap file if the amount of memory does not exceed 512MB.

macmpi:
Dunno if completely related, but can we make sure activity intensive logs & caches (included `spop`'s in `/run/shm`, unsure where Volspotconnect cache is) stay in an allocated real RAM space, even in 512MB situation.
In particular Spotify is known to be quite intensive and may alter flash.

PR #124, opened 2017-01-14
Use mirrordirector.raspbian.org for Pi builds

macmpi:

PR #122, opened 2017-01-14
Use repo redirector

macmpi:
consistency with other recepies

PR #130, opened 2017-01-15
Update recipes sources repo

macmpi:
replaces https://github.com/volumio/Build/pull/122 and https://github.com/volumio/Build/pull/124


gkkpch:
I think it is a good thing to have the redir in the recipes, I used that for the armv7 and armv8 too.
However, you have to realize it is not guaranteed to work all the time. I had various issues with it.
In those cases you need to be patient and try later, or temporarily change it back to a local mirror.

macmpi:
Yes wanted to make is consistent.
While at it, now `apt-transport-https` is included in all recipes (except `arm-minimal`), wouldn't it make sense to **https** all repos?

volumio:
I don't see a real benefit...
Taking into account what Gé reported earlier about redir, should we revert those commits?

macmpi:
I guess keeping redirectors is good practice in a distributed development environment as build may happen in different geographies.
https becomes a necessary good practice to ensure downloads are from legitimate sources.

PR #134, opened 2017-01-22
Fix low-entropy condition (haveged) and add nl80211 driver to wpa_sup…

gkkpch:
…plicant


macmpi:
Yes, `haveged` helps indeed on low entropy systems.
Mind the typo [here](https://github.com/volumio/Build/pull/134/commits/b21b689b0b18d679bece7c4f73882fe9e2853e98#diff-ca982d185cd1421ab96b9efb00457184R34)

macmpi:
nl80211 is indeed the most widely used one (I added it in hostpad.conf for std drivers).
However I'm unsure Edimax properly supports it (one reason why Edimax dongle has specific `hostapd.conf`): did you have any chance to check?

gkkpch:
@macmpi Thanks for the typo info!
I don't have an Edimax, but I presume this is still ok. 
Another 8192cu based dongle here appears to be working fine.

macmpi:
You're welcome, I'll let you submit the fix since it has has been merged already.

Incidentally had some interesting read [here](https://wiki.archlinux.org/index.php/WPA_supplicant#Problem_with_mounted_network_shares_.28cifs.29_and_shutdown) about related issue with NAS unmount delaying shutdown: that's something we experience sometimes in Volumio.
What do you think about the described workaround? Shall we include that?

macmpi:
about low-entropy solved with `haveged`: just came across one might need to consider [rng-tools](https://wiki.archlinux.org/index.php/Rng-tools) instead for increased performances (or in complement):
> The rng-tools is a set of utilities related to random number generation in kernel. The main program is rngd, a daemon developed to check and feed random data from hardware device to kernel entropy pool.
> 
> This is mainly useful to increase the quantity of entropy in kernel to make /dev/random faster. By default, /dev/random is very slow since it only collects entropy from device drivers and other (slow) sources. rngd allows the use of faster entropy sources, mainly hardware random number generators (TRNG), present in modern hardware like recent AMD/Intel processors, Via Nano or even Raspberry Pi.

It [just got integrated in Raspbian](https://github.com/raspberrypi/linux/issues/2799#issuecomment-478124601) base to increase low-entropy performance

volumio:
@macmpi interesting. Could you come up with a PR to substitute haveged with rng-tools?

macmpi:
Replacing `haveged` by `rng-tools` in each platform recipe should make the trick, but I'm actually not sure of the exact situation/use-case/issue that originally revealed the entropy issue.
We'd need to test the same use-case with `rng-tools` to ensure there is no regression.
Maybe @gkkpch remembers?

PR #137, opened 2017-01-29
boot mgr fixed

chrismade:
some minor fixes in build.sh and scripts/bpipro*.sh - can now directly create a bootable Banana-Pi Pro image without manual fine tuning - kept @gkkpch 's kernel for this commit


volumio:
Great thanks! Build now and see...

PR #154, opened 2017-02-28
add MrEngman drivers for armv6 and armv7

macmpi:
tentative: not fully tested as I did not build complete image.
Any comment welcome


volumio:
Clever one... Is it tested?

macmpi:
won't work, canceling.

PR #158, opened 2017-03-03
Cubox: Linux header files added to /usr/include

gkkpch:


volumio:
Gé, without including it ( I assume this considerably increase the image size) Can we just link to it somewhere or script the installation so it happens on demand?

gkkpch:
Actually, it is hardly noticeable. 
Only 839Kb increase in zipped image size when stored as a tarball in /usr, which is approx. 3 per mille.

volumio:
Oh, tiny! Let's merge, good work gé

PR #161, opened 2017-03-07
Use proper 4.4 branch for rpi-update

macmpi:
rpi-update (default) **master** branch is now on 4.9.y
Use **stable** branch while volumio sits on 4.4.y

### DIFF
--- a/scripts/raspberryconfig.sh
+++ b/scripts/raspberryconfig.sh
@@ -61,10 +61,11 @@ mkdir /lib/modules
 KERNEL_VERSION="4.4.9"
 KERNEL_REV="884"
 
-echo y | SKIP_BACKUP=1 rpi-update 15ffab5493d74b12194e6bfc5bbb1c0f71140155
+# using rpi-update stable branch for 4.4.y as master is now on 4.9.y
+echo y | SKIP_BACKUP=1 BRANCH=stable rpi-update 15ffab5493d74b12194e6bfc5bbb1c0f71140155
 
 echo "Updating ELF"
-echo y | SKIP_KERNEL=1 rpi-update
+echo y | SKIP_KERNEL=1 BRANCH=stable rpi-update
 
 echo "Adding PI3 Wireless firmware"
 wget http://repo.volumio.org/Volumio2/wireless-firmwares/brcmfmac43430-sdio.txt -P /lib/firmware/brcm/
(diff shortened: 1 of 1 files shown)


volumio:
No way, we can't change kernel now. Simply because people will need to reflash and can't use the updater. 
We're preparing a new updater that will update the kernel. Until we have it, we stay in this kernel.
Plus, we should ALWAYS use a kernel commit, since we need to keep it consistent across releases. Last thing, before releasing a kernel we need to do extensive testing to see if everything (i2s DACs mainly, works as expected)

macmpi:
I think you missed the point I made.
It is NOT about changing the kernel indeed.
But it is a fact that rpi-update master is now on 4.9 (since a few weeks).
So if we want to avoid bizarre updates in the build process with firmware files coming from 4.9, then we are better explicitly pointing to the right branch now.

volumio:
Sorry I misunderstood indeed. 
So, basically this will just stick the firmware to the stable branch, correct ?

macmpi:
Yes.
This will by the way help supporting newer PiZero W.
Firmware got updated March 3rd, right after v2.114 build !...

volumio:
Mmm guess its too late then, lol 
http://updates.volumio.org/pi/volumio/2.118/volumio-2.118-2017-03-07-pi.img.zip

macmpi:
you may want to check firmware files, as they probably come from 4.9 branch!...

volumio:
Yep, I will check them, and check this commit and eventually merge...

macmpi:
Hopefully they are the same...
But pointing to proper branch is more future proof until you move to next kernel: actually it's a good lesson learn, let's make explicit reference to branches when possible, as master moves.

macmpi:
Firmware files are definitely not the same between the 2 branches.
Made a build with the changes, and it has proper 4.4 firmware files.

Still, something strange, newer PiZero W .dtb does not come... (in either case BTW: not present in 2.118-2017-03-07-pi.img.zip either which [does not work on PiZero W](https://volumio.org/forum/with-new-rpi-zero-t6050.html#p29551).

macmpi:
Actually firmware files from 4.4 (stable) and 4.9 (master) do not look the same...
Built an image with the proposed change, and files do come from 4.4 branch, but MD5 is different from stable ones: I'm quite puzzled.

Another weird thing, in my image, and the one you built, new PiZero W .dtb file is not present!...

volumio:
OT: Could you help me on this one? I think you might have something up your sleeve! 
https://volumio.org/forum/installing-the-touch-display-plugin-t6032.html#p29602

macmpi:
hum... 2.114 probably had the 4.9 firmware mix already. Let me check if the plugin installs fine on my recent build...(PS, can't find how to view the build log)

macmpi:
@volumio did you have a chance to check this?
I'm running my own Build image with it for few days, and it looks fine: it has consistent 4.4 files.

I don't think we want to keep a mixed bag of 4.4 and 4.9 files in the build in upcoming release images as it is today with 2.118 (and maybe 2.114): this may produce quite unexpected outcomes at some point.

PR #163, opened 2017-03-08
protect /dev/shm

macmpi:
just a good practice
https://www.cyberciti.biz/faq/linux-add-nodev-nosuid-noexec-options-to-temporary-storage-partitions/

### DIFF
--- a/scripts/armbianconfig.sh
+++ b/scripts/armbianconfig.sh
@@ -14,7 +14,7 @@ tmpfs   /var/log                tmpfs   size=20M,nodev,uid=1000,mode=0777,gid=4,
 tmpfs   /var/spool/cups         tmpfs   defaults,noatime,mode=0755 0 0
 tmpfs   /var/spool/cups/tmp     tmpfs   defaults,noatime,mode=0755 0 0
 tmpfs   /tmp                    tmpfs   defaults,noatime,mode=0755 0 0
-tmpfs   /dev/shm                tmpfs   defaults        0 0
+tmpfs   /dev/shm                tmpfs   defaults,nosuid,noexec,nodev        0 0
 " > /etc/fstab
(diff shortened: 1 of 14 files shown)


macmpi:
Running my Pi for few days already with this fstab change, and all is great.
Maybe I should propagate this to other platform scripts too, so that the PR is consistent across the board?

volumio:
Yes please extend it to other platforms...

macmpi:
Done. Amazing number of supported platforms indeed :)
only x86 missing, but I guess it's somewhere else

gkkpch:
Yes, x86 works with a template file, I'll fix it later tonight

On 14 Mar 2017, at 15:46, macmpi <notifications@github.com> wrote:

Done. Amazing number of supported platforms indeed :)
only x86 missing, but I guess it's somewhere else

gkkpch:
The x86 version is not using shared memory, I also doubt it would have any noticable advantage in this case.

volumio:
Gé, do you think it might bring regressions for arm based platforms?

gkkpch:
Difficult to judge as I don't know which volumio system parts are heavily using share memory to communicate.
Only regression testing can tell and that would need to be done for at least a few of them.

On 14 Mar 2017, at 21:01, Volumio <notifications@github.com> wrote:

Gé, do you think it might bring regressions for arm based platforms?

volumio:
We usually do write some stuff in /tmp which is mounted on RAM, some files we write on /tmp need to be executable

macmpi:
Yes and it is still the case with running the change: one can still execute things in `/tmp` (just confirmed again on my Pi)
Only `/dev/shm` is protected

volumio:
Should be fine then. Merging

PR #164, opened 2017-03-11
Issue 143

earlchew:
I've run this code for a while now without issue, so it's probably a good a time to consider merging.


macmpi:
Was it just run on a target or built with build system?
Did not look in details, but have you taken care of linking newly created (or renamed) services to proper final systemd location in Build system script, as per [this](https://github.com/volumio/Build/blob/master/scripts/volumioconfig.sh#L413 for instance?)

volumio:
Sorry for not merging. Honestly I think this introduces too many new stuff in a single bunch. I would rather do a progressive approach where we change stuff little by little. What do you think?

earlchew:
@volumio I'm ok to rework into a series of smaller pull requests.

earlchew:
@macmpi I did not run this from a clean build. I think I should probably do so, meaning that I will have to take time out to construct a complete build system.

macmpi:
First page of Build github has very simple howto: it is really that simple & quick.
I would not recommend installing docker package at-all though, as in my case it [altered apt-get](https://volumio.org/forum/x86-build-script-error-t6161-10.html#p30309) operations during build...

earlchew:
@macmpi Thanks for the hints. Perhaps you'll still be available to help with testing?

macmpi:
Sure, I have a keen interest on all those system/network stuff.
I'll swap-in some SD with some image you may provide: FYI, I'm on PiZero and USB wifi dongle

PR #165, opened 2017-03-13
Pi: Block unwanted firmware & kernel updates

macmpi:
This holds critical packages to prevent direct or indirect (through dependencies) installation of libraspberrypi0, raspberrypi-bootloader, raspberrypi-kernel which critically upgrades kernel & firmware files.
This may happen when installing packages such as chromium, sense-hat, python-picamera...

This change has been tested in a new build: successfully blocked such kind of installs.
Also confirmed newer Touchscreen plugin still installs fine after this change.


volumio:
Thanks!

PR #166, opened 2017-03-13
Initial support for PiZero W wireless

macmpi:
Add necessary bcm2708-rpi-0-w.dtb from 4.4 stable branch
Probably won't be necessary anymore when Kernel is updated, so it's a conditional install for now

Reported working by several users in [Forum](https://volumio.org/forum/with-new-rpi-zero-t6050-20.html#p29932).
(note: uses same chipset than Pi3 so no need for new wireless firmware)


volumio:
Well done!

PR #168, opened 2017-03-14
Pi: add pi3-disable-wifi.dtbo

macmpi:
This allows users to completely turn off pi3 wireless chip (even radio), by setting `dtoverlay=pi3-disable-wifi` in `/boot/config.txt`

Tested working in this [Forum thread](https://volumio.org/forum/rpi-and-wifi-dongle-t6099-10.html#p29954)

Note that as per current issue in https://github.com/volumio/Volumio2/issues/946 , i2s_dacs/index.js always overwrites existing content of `/boot/config.txt`.
This compromises any changes made into `/boot/config.txt` by users and/or possibly required by some plugins.
That issue needs a fix, so that such features can be consistently used.

### DIFF
--- a/scripts/raspberryconfig.sh
+++ b/scripts/raspberryconfig.sh
@@ -83,7 +83,7 @@ if [ "$KERNEL_VERSION" = "4.4.9" ]; then       # probably won't be necessary in 
 echo "Adding initial support for PiZero W wireless on 4.4.9 kernel"
 wget -P /boot/. https://github.com/raspberrypi/firmware/raw/stable/boot/bcm2708-rpi-0-w.dtb
 echo "Adding support for dtoverlay=pi3-disable-wifi on 4.4.9 kernel"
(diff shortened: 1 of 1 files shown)


macmpi:
Thanks.
Can you fix the [i2s_dacs/index.js](https://github.com/volumio/Volumio2/issues/946) issue that rewrites `/boot/config.txt`  as a whole instead of just search/replacing necessary character string, as this potentially kills any change made by user to that file?

xipmix:
I have been thinking about this a bit, it looks like there are a few sources for updates to /boot/config.txt which makes this tricky to get right. 

Is it possible for there to be more than one ```dtoverlay=``` line in a /boot/config.txt file?

macmpi:
@xipmix thanks for looking at that! Can we discuss this in the original opened issue https://github.com/volumio/Volumio2/issues/946 ?

PR #167, opened 2017-03-14
udev rules to prioritize network interfaces

macmpi:
This PR replaces [Pi3-specific driver blacklist-based filter](https://github.com/volumio/Build/issues/150) by udev rule.
It brings several benefits:
- avoids unwanted brcmfmac/brcmutil driver loads on all Pi platforms
- prioritizes plugged-in external USB network interfaces over built-in (wlan & eth)
- generic for all platforms (+added customization for Pi built-in eth which **is** on USB)

This fix has been tested with several devices and dongles.

PR #169, opened 2017-03-16
Add firmware for Atheros AR9170 wifi USB dongle

macmpi:
[Reported working](https://github.com/volumio/Build/pull/167#issuecomment-286980405) by @chsims1 with [that addition](https://github.com/volumio/Build/pull/167#issuecomment-286470758) on Pi platform (client to Home wifi AP only, not Volumio Hotspot at this point).

Actual carl9170 driver is in kernel, but carl9170 firmware is in `firmware-linux-free` (rather than in `firmware-atheros`! )

Shall I also add package for other builds?

### DIFF
--- a/recipes/arm-dev.conf
+++ b/recipes/arm-dev.conf
@@ -54,7 +54,7 @@ keyring=debian-archive-keyring
 suite=jessie
 
 [Firmware]
-packages=firmware-atheros firmware-ralink firmware-realtek
+packages=firmware-atheros firmware-ralink firmware-realtek firmware-brcm80211 firmware-linux-free
 source=http://mirrordirector.raspbian.org/raspbian
 keyring=debian-archive-keyring
 components=main non-free
--- a/recipes/arm.conf
+++ b/recipes/arm.conf
@@ -48,7 +48,7 @@ keyring=debian-archive-keyring
 suite=jessie
 
 [Firmware]
-packages=firmware-atheros firmware-ralink firmware-realtek firmware-brcm80211
+packages=firmware-atheros firmware-ralink firmware-realtek firmware-brcm80211 firmware-linux-free
 source=http://mirrordirector.raspbian.org/raspbian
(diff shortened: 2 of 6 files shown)


macmpi:
for info, package [content](https://packages.debian.org/sid/firmware-linux-free)

macmpi:
Also added for all arm except minimal (does not have any firmware)
x86 already has it through `firmware-linux`, and x86-dev does not seem to have any firmware.

macmpi:
Some noteworthy [driver limitations](http://linuxwireless.org/en/users/Drivers/carl9170/).
In particular, one may want to enable `nohwcrypt=1` because of potentially failing HW encryption (is a classical Atheros thing, like for ath9k...):
`echo "options carl9170 nohwcrypt=1" | sudo tee /etc/modprobe.d/carl9170.conf`

Will add to Build if confirmed WPA/WPA2 issues are reported:  (@chsims1 ?)
- fails to join WPA2 Home AP without this option (restart after setting option)
- fails to create Volumio Hotspot without this option (restart after setting option)

volumio:
Ok I'll wait to merge until you got the nohwcrypt confirmed. Is that ok?

macmpi:
@chsims1  may you check if `/etc/modprobe.d/carl9170.conf` I created on your Pi3 provides any benefit with connecting your Atheros dongle, either to HomeAP or as a Hotspot?
It may fix some authentication errors if any...
You may rename/remove to check if any difference.
If it adds benefit, I will add to build.
Thanks.

chsims1:
That file is not there.  Anyhow, the Atheros dongle works both as connection to home AP and as a hotspot.

macmpi:
Ok thanks (I may have removed the file at some point when doing clean-up Saturday evening).

Good, so @volumio can probably merge PR as-is, without the additional file for now.
At least it's tracked here if anyone have connections issues/drops in the future (may very well be Home AP setup dependent).

PR #170, opened 2017-03-16
Pi: Fixing MrEngman repo

macmpi:
Repo has changed location since recent Dropbox [policy change](https://www.raspberrypi.org/forums/viewtopic.php?p=1131942#p1131942)
Put base repo in variable in case of further changes to be more future-proof.

Tested recomposed URLs seems good now.


volumio:
Thanks! Do they work with our kernel version? Did you try a build?

macmpi:
Nope did not try a new build yet, but those are exactly the same files as previous: only base URL has changed, as they were essentially copied-over from one web site to another one.
Files download OK

volumio:
Ok if files download are ok, I merge

PR #171, opened 2017-03-23
Add Aoide-DACs support.

howardqiao:
Add Aoide-DACs support.


volumio:
Hi, thanks for the PR!
Before integrating we need to make sure that they are GPL and we're GPL compiant:

Please confirm that:

The code submitted is released under GPL (indicate the version)
Please show where sources for the binaries can be found
You will take all responsabilities and obligations concerned by the distribution of the binaries you provided. This includes making the sources available to everyone requiring them.
You agree to free Volumio and INTUITU di Michelangelo Guarise from any responsability deriving from distribution of the binaries you provided, and to provide eventual indemnifycation for damages

Reply "I agree", along with your name to agree

volumio:
Sorry, not merging a change of repository

gkkpch:
Why are we changing the kernel and removing all Allo DAC modules?

macmpi:
Yes very confusing indeed.
@howardqiao are your trying to include the same thing as here https://github.com/raspberrypi/linux/pull/1559#issuecomment-232751431 ?

volumio:
@howardqiao There's no way I will merge this. To get your DAC up and running, please just provide a zip file for the 2 kernels you see in raspberryconfig.sh (better if via email) and agree to the terms in the second message. So I am closing this one, please open another PR doing what I've asked, since its the only way we can accept your work. Thanks

PR #174, opened 2017-03-29
Pi: prepare to test new kernels

macmpi:
case switch to allow testing Build with different kernel versions


volumio:
So, this won't change anything when merged, but we can add a case statement for a newer kernel and build it if we change kernel number, correct?

macmpi:
yep

volumio:
Ok, next step is to branch it with a newer kernel . How could we determine the best version for 4.9 ?

macmpi:
I've been tracking [this](https://www.raspberrypi.org/forums/viewtopic.php?f=29&t=167934) for a while, along with kernel repo changes.
Seems current one (4.9.17) is quite mature, with some important fixes in.
Many of key DAC overlays seem there since 4.9.11: you may want to get feedback from your DAC vendors contacts if they are fine on 4.9.x?
LibreElec is on 4.9.13 but I'd rather go north of it.

Issue #172, opened 2017-03-24
/etc/mtab link incorrect?


macmpi:
Trying to get x86 image working on my [mid-2010 MBP](http://www.everymac.com/systems/apple/macbook_pro/specs/macbook-pro-core-2-duo-2.4-aluminum-13-mid-2010-unibody-specs.html) (some earlier Volumio2 image did work) or under Virtualbox (converted image from .img into .vdi), and undergoing some similar first boot issues ending into recovery mode.

I'm getting over some `dmesg` debugging, and incidentally noted a reported issue with improper `mtab` (see attached screenshot).

`mtab` related stuff [in Build](https://github.com/volumio/Build/search?utf8=%E2%9C%93&q=mtab&type=) are only referencing mkinitramfs scripts, and in that case symbolic link seems incorrect compared to what it is supposed to be in `systemd` context, as per displayed error log in screenshot:
`ln -sf /proc/self/mounts /etc/mtab`
Is that intended, or shall we fix it ?

![screen shot 2017-03-24 at 12 09 40](https://cloud.githubusercontent.com/assets/16296055/24291832/ef6e9c04-108a-11e7-9ace-a816dbd6fd94.png)

macmpi:
no more issue with 2.129

PR #177, opened 2017-03-29
fix mtab symlink 

macmpi:
as per https://github.com/volumio/Build/issues/172

### DIFF
--- a/scripts/initramfs/mkinitramfs-custom-single.sh
+++ b/scripts/initramfs/mkinitramfs-custom-single.sh
@@ -300,7 +300,7 @@ fi
 
 # fstab and mtab
 touch "${DESTDIR}/etc/fstab"
-ln -s /proc/mounts "${DESTDIR}/etc/mtab"
+ln -sf /proc/self/mounts "${DESTDIR}/etc/mtab"
 
 # module-init-tools
 copy_exec /sbin/modprobe /sbin
--- a/scripts/initramfs/mkinitramfs-custom.sh
+++ b/scripts/initramfs/mkinitramfs-custom.sh
@@ -260,7 +260,7 @@ fi
 
 # fstab and mtab
 touch "${DESTDIR}/etc/fstab"
-ln -s /proc/mounts "${DESTDIR}/etc/mtab"
+ln -sf /proc/self/mounts "${DESTDIR}/etc/mtab"
 
 # module-init-tools
 copy_exec /sbin/modprobe /sbin


volumio:
What's the purpose of this one?

macmpi:
error report in referred issue https://github.com/volumio/Build/issues/172 (system can't boot & explicit error message), and consistent with [this](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=494001) and many other similar sources.

volumio:
@gkkpch what do you think?

gkkpch:
I will have a look this weekend, i'm not sure if mtab really matters. I'm not aware we configured/ changed it. Not used in initramfs, so sometging else must be causing the issue.

macmpi:
It seems some Debian release with **pre 2.27.1** `util-linux` [had a strict enforcement](https://github.com/systemd/systemd/issues/1495) for this `mtab` link and `systemd` would freeze boot.
This very much seems to be the symptom I experience there https://github.com/volumio/Build/issues/172, as built-in `util-linux` in my x86 image is **2.25.2-6**

So we might need more recent package and/or adapt the link accordingly (which seems a good practice in many references).

gkkpch:
I'm not convinced yet, that this causes the problem. AFAIK we did not have any boot problems with x86 yet, other than missing ata/sata/pata modules in initramfs, forcing the boot process into a kernel panic. Your boot problem occurs a lot later and that is the part I need to investigate. But, does your patch fix the issue?

macmpi:
Did not have a chance to build an x86 image yet...

But you're right I most probably face a more fundamental issue: no filesystem seems mounted at all when in emergency mode at end of boot (so I can't even save dmesg to send it!).
`df: cannot read table of mounted file systems`
So it seems initramfs goes through, not mounting stuff properly, but still passes to system which gets into emergency mode...
Dunno what I can do from there to debug...
screen shot 2017-03-30 at 16 06 30

gkkpch:
Plus the symbolic links you are referring to, are for the initramfs system. They get abandoned just before doing the switch_root and loading linux. Let me look at your issue, should have time again tomorrow

gkkpch:
@macmpi had a closer look at your screen shot, looks like your rootfs (overlayfs from squash file and data partition) has difficulties. Just to make sure we are using the same build, can you remember which version this was? If not, you might get a glimpse while it is running initramfs, comes out as one of the last echos before switching rootfs. I will then burn the same image and test.

gkkpch:
I had a closer look, the point where
"    1.854468] squashfs: version 4.0 (2009/01/31) Phillip Lougher" 
was logged, was where initramfs started.
Would you be able to do a journalctl -b instead of dmesg?
And am I right that this is from a virtualbox?
In that case I'm not sure if I want to put much time in it ;)
I burned the image you mentioned, booted from it on my notebook without issues...

macmpi:
Thanks for looking indeed.
Yes, it is on Virtualbox currently, but as mentioned in initial issue, I had the same problem on my MacBook initially (and a PC notebook).
That's the reason why I started looking into Virtualbox which is much more convenient to start, reboot, while using my principal machine (same Macbook) normally. Some earlier image (can't remember which now) did work on all these environment BTW.
Here is [journalctl -b](https://pastebin.com/EMLns1Pf) report.

gkkpch:
I'm sure now it is unrelated to  /etc/mtab. We need to find out what goes wrong before booting linux. 
What notebook (brand and model) were you testing on? Macbook has not been confirmed working, we had other issues with Mac and I'm not sure we solved them yet. In the end all boiled down to disk access and missing/ failing modules

macmpi:
I can confirm some earlier version did work on my [mid-2010 MacBook Pro](http://www.everymac.com/systems/apple/macbook_pro/specs/macbook-pro-core-2-duo-2.4-aluminum-13-mid-2010-unibody-specs.html).
Other working machine was an old ASUS eeePC 1005HA, third was VirtualBox.
None work with this image.

BTW, do you maintain a list of confirmed machines across versions somewhere?

gkkpch:
I find communicating via github extremely complicated (not always near a pc) and off-putting, can we not find another way (e.g. our whatsapp group)?

macmpi:
https://gitter.im/apps mobile apps may be an option?

macmpi:
reading [this](https://volumio.org/forum/compatible-x86-machines-t6301.html#p31029):

> Brand: VirtualBox 5.1.18
> Booting from: locally built X86 image .vmdk
> Confirmed image version: newer than 2.129

Did you find a fix for Virtualbox? Nice!
Hopefully it will also fix my other machines then...

Maybe you can save building `.vmdk`, and instead convert `.img` on host Virtualbox machine with:
`VBoxManage convertfromraw --format VDI [filename].img [filename].vdi`
usually works pretty well (--format VDI not required but makes it more obvious)

gkkpch:
Sorry, no fix for virtualbox because there was nothing to fix. I use the x86 build script as is and then use the vmdk file to create a new virtual machine as always. Never had a problem with it, I have done the initrd, the uefi an legacy boot development with it. Don't convert the build image, just use the VirtualBox options. Perhaps that solves your problems. I cannot reproduce them, sorry.

macmpi:
ok, just tried with 2.129 under Virtualbox (converted release .img image), and all is good.
Was probably a 2.115-specific issue in the end...
Will test on real HW as soon as possible and report setup in Forum.
Thanks.
I'll close PR once other tests done.

macmpi:
Closing: 2.129 OK with Virtulabox, EeePC 105HA and [Macbook](https://volumio.org/forum/compatible-x86-machines-t6301-10.html#p31170) (albeit no wifi probably because of missing broadcom **wl** driver).

gkkpch:
well, if Broadcom does not make the driver available, then so be it. This is a typical case for people to add their suppliers own propriety stuff (usually from that famous extra CD), nothing to do with volumio. The x86 world is a difficult one, we can only support the generic stuff as supported by Debian and some well known repos.

PR #178, opened 2017-04-02
Do not install custom mpd package on armv7/armv8

gkkpch:


macmpi:
Is this intended for [this](https://volumio.org/forum/pine-large-library-nas-t6152-10.html#p31007) fix?
Isn't it that we need a "custom mpd" (same sources than for arm) but actually built for arm7 and arm8?
Would not this also apply for all other packages installed in that section of `volumioconfig.sh` which are pulled from volumio repo (upmpdcli & libs, shairport, etc...)?

gkkpch:
No, also for Odroid C1/C2, Cubox, Sparky, Udoo QDL and Udoo Neo. We don't necessarily need a custom one for armv7/armv8. As for the other packages, we do not suffer the alignment issues, so no action for the time being.

volumio:
I think we should do it differently: we want to have 0.20.4 so we just need to differentiate beetween armv6 and armv7. We can take debian experimental package for armv7, while for armv6 we will have a compiled one. What do you think?

gkkpch:
Installing 20.4 from experimental is not the only thing, at the same time a lot of libraries and further dependencies are needed from sid (unstable). I found a way to do it after a lot of experimenting, adding sid and experimental to sources.list at the right time in the device config script (tested with a C2). However, when doing this in volumioconfig.sh half the rootfs gets replaced, an absolute no-go.  But even with the device config, the combination of experimental and sid is quite risky if you ask me.

macmpi:
I'm not sure what it the primary reason for these "custom" packages (mpd, upnpdcli, etc): non-availability on std repo, or custom-patched souce-code?

In any-case, it's probably a good idea to build them all with the most scrupulously matched dependencies, and eliminate such potential bugs risk for good: this is probably much more future proof.
Indeed, we may get very unpredictable results by using binaries on system with unmatched dependencies: you're actually lucky when you get segfaults because then bug symptom is obvious.
But quite often things just get erratic, and it's extremely hard to figure-out and find root cause (lesson learned the hard way)...

macmpi:
@volumio , how do you wish to better segment installs inside `volumioconfig.sh`, to separate arm/armv7/armv8?
Currently we treat all ARM variants together with [this test](https://github.com/volumio/Build/blob/master/scripts/volumioconfig.sh#L108) on qemu.
Or shall we switch gear and do all binary installs within each specific `platformconfig.sh`?
Thanks for suggestion.

PR #179, opened 2017-04-07
Pi: GPIOs configuration clean-up

macmpi:
This PR intends to refresh GPIO setup on Pi, to make it more future-proof and aligned with jessie:
- update udev rules to current jessie distribution, properly allowing userspace use of GPIOs
- leverage current [wiringpi](http://wiringpi.com/) from raspberrypi.org repo (BTW seems only used for [this](https://github.com/volumio/Volumio2/blob/master/app/plugins/system_controller/i2s_dacs/scripts/iqamp-unmute.sh), we could save it eventually)
- remove defunct (and seemingly unused) [gpio-admin](https://github.com/quick2wire/quick2wire-gpio-admin)
- cosmetic tuning of other I/O groups

Has been tested by building a current ROM image (kernel 4.4.9): still operated correctly with GPIO-Buttons plugin.

### DIFF
--- a/scripts/raspberryconfig.sh
+++ b/scripts/raspberryconfig.sh
@@ -113,11 +113,12 @@ wget http://repo.volumio.org/Volumio2/wireless-firmwares/brcmfmac43143.bin -P /l
 echo "Installing WiringPi from Raspberrypi.org Repo"
 apt-get -y install wiringpi
 
-echo "adding gpio group and permissions"
+echo "adding gpio & spi group and permissions"
 groupadd -f --system gpio
+groupadd -f --system spi
 
-echo "adding volumio to gpio group"
-sudo adduser volumio gpio
+echo "adding volumio to gpio group and al"
+usermod -a -G gpio,i2c,spi,input volumio
 
 echo "Use up-to-date jessie rules for gpio & al."
 read -rd '' Rule_String <<"EOF"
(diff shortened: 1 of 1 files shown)


scripts/raspberryconfig.sh
```diff
@@ -110,27 +110,52 @@ wget http://repo.volumio.org/Volumio2/wireless-firmwares/brcmfmac43143.bin -P /l
 #dpkg -i /raspi/raspi-config_20151019_all.deb
 #rm -Rf /raspi
 
-echo "Installing WiringPi"
-wget http://repo.volumio.org/Volumio2/Binaries/wiringpi_2.24_armhf.deb
-dpkg -i wiringpi_2.24_armhf.deb
-rm /wiringpi_2.24_armhf.deb
+echo "Installing WiringPi from Raspberrypi.org Repo"
+apt-get -y install wiringpi
```
volumio: It might be safer to download current version of wiringpi and place the .deb on Volumio's repo. This way we will avoid some breaking changes in future versions. What do you think?

macmpi: I think `wiringpi` is a very-well supported component, and availability in `raspberrypi.org` repo is very safe for longterm (mostly similar confidence than with Raspbian) : this should ensure maximum compatibility & bug-fixes warranty.
I consider being lucky to have spotted issue with specific volumio repo`wiringpi_2.24` and new kernel. The more we lock-in to specific versions in volumio repo, the more we get exposed to such upgrade issues I think.
We may want to only have very specific tweaked software in volumio repo: the fewer there are, the more manageable it is.

PR #181, opened 2017-04-09
Hostapd-edimax use: improved compatibility

macmpi:
Custom `hostapd-edimax` is only launched if the loaded 8192cu module driver does NOT support `cfg80211` (as with pre 4.4.13 kernel)
Indeed, since 4.4.13 Pi kernel release, 8192cu module driver has been enhanced (without changing its version number) to include `cfg80211` support, and therefore now requires standard `hostapd`.

Note: this fix does not alter operations on current 4.4.9-based Volumio, as it indeed keeps using `hostapd-edimax` as previously.
It will also work with other ARM platform according to the 8192cu driver they may use.

### DIFF
--- a/volumio/bin/hotspot.sh
+++ b/volumio/bin/hotspot.sh
@@ -2,10 +2,10 @@
 
 case "$1" in
 'start')
-DRIVER=`/sbin/ethtool -i wlan0 | grep driver | awk -F": " '{print $2}'`
+MODULE=$(basename $(readlink /sys/class/net/wlan0/device/driver/module))
 ARCH=`/usr/bin/dpkg --print-architecture`
 
-if [ $DRIVER = "rtl8192cu" -a $ARCH = "armhf" ] ; then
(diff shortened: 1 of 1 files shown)


volumio:
So basically you're checking if the current module supports cfg80211, if not you launch hostapd edimax?

macmpi:
Yes, only if wlan0 actually uses related `8192cu` module (similar as before, although it does test the actual module name, rather than the driver name (`rtl8192cu`) as previously. 
Indeed, `rtl8192cu` may be implemented by some other driver in the future, so the special `hostapd-edimax` case is really made specific to `8192cu`.

For other drivers, we do leverage specific `cfg80211` check (other part of the `if` will indeed be `false`) since anyway hostapd-edimax **is only relevant to** `8192cu` driver.
If you want we could do sth for other non-`cfg80211` compatible drivers, and send a warning log message and fail gracefully.
But then this is more a feature addition, rather than just compatibility fix.
Maybe for later?

volumio:
Yep, maybe later...

PR #187, opened 2017-04-30
Pi: bump next kernel test case to 4.9.25

macmpi:
4.9.25 includes various improvements over 4.9.20 (networking, i2s,...)
As always, this change does NOT affect standard build of Volumio on 4.4.9


macmpi:
was thinking: maybe [these guys](https://volumio.org/forum/hifiberry-digi-pro-and-volumio-t4740-40.html) would provide a good motivated test-crew for this...

volumio:
Indeed... Give me few days and I'll make a dev build for them

PR #191, opened 2017-05-04
Pi: set firmware commit reference

macmpi:
note: BRANCH is actually ignored by rpi-update when a commit# is set
Just keeping it for now still, as at some point (when support 4.4.9 depreciated) we may sync with "stable" branch?

### DIFF
--- a/scripts/raspberryconfig.sh
+++ b/scripts/raspberryconfig.sh
@@ -66,11 +66,13 @@ case $KERNEL_VERSION in
       KERNEL_REV="884"
       KERNEL_BRANCH="stable"
       KERNEL_COMMIT="15ffab5493d74b12194e6bfc5bbb1c0f71140155"
+      FIRMWARE_COMMIT="9108b7f712f78cbefe45891bfa852d9347989529"
       ;; 
     "4.9.25")
       KERNEL_REV="994"
       KERNEL_BRANCH="master"
       KERNEL_COMMIT="a86bfee5b47a74c13056997f1e4d8b9d8090b398"
+      FIRMWARE_COMMIT=$KERNEL_COMMIT
       ;; 
 esac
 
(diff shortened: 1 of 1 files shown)


volumio:
Yes, we will sync with the kernel in the future, it will be the best thing to do. Thanks, testing and merging

volumio:
It works on the only PI3 I have with proven working wireless. 
The other 2 PI3's with flaky wireless don't work. I am now trying with 2.129

macmpi:
If you md5 compare some `/boot` firmware files within original 2.129 and 2.141 images, are they different?

If you generate a new 2.141 (same volumio2 FE/BE commit#) with the updated build, do you obtain the same firmware files in `/boot`?

volumio:
It seems that I have at least 3 broken PI3s (wireless). I have only one working here, which works great with 2.129 and latest build with your PR.
So I would say we could merge this one, as it seems an hardware issue on my end. 

Do you still want me to confront /boot ? it might be a long task...

macmpi:
Are the "broken" Pi3 of a recent batch purchases? (or at least more recent than the working one)?

You may want to test-build under 4.9.25 and see if "broken Pi3" work then: this would confirm the most up-to-date firmware/kernel fixes the issue definitely.

volumio:
I would say the other way around, the working one has a "brighter and greener" silk screen. I will now build a recent kernel and see

macmpi:
Did new kernel bring any benefit with the other "broken" Pi3s?

We got caught twice by such unexpected branch changes very recently: dunno if we can subscribe to some pre-announcement list to avoid such discomfort.

When we move to new release family, it might be worth considering doing kernel/firmware installs with `apt-get raspberrypi-bootloader raspberrypi-kernel` rather than `rpi-update`: this ensures current pinned as "stable" components are installed. Maybe safer when we stick to a same release family (they may change stable family at a point, but is is less often at least!).
We could still `hold` such chosen version (you may keep a copy on your repo), and unhold/hold when doing updates.
What do you think?

PR #192, opened 2017-05-06
Pi: give-up on KERNEL_BRANCH references

macmpi:
Got confirmation from https://github.com/Hexxeh/rpi-firmware/ repo maintainers there's not plan to notify on drastic branches changes. Therefore we are safer to just stick to commit# while we use `rpi-update`.


volumio:
Ok, that's wise. Hopefully we'll avoid surprises...

PR #197, opened 2017-05-31
Pi: Adding Allo Firmware independently of Kernel Version

macmpi:
Only modules and dtbo may be kernel-dependant.
Firmware always need to be installed.

### DIFF
--- a/scripts/raspberryconfig.sh
+++ b/scripts/raspberryconfig.sh
@@ -284,20 +284,9 @@ rm /patch
 
 if [ "$PATCH" = "volumio" ]; then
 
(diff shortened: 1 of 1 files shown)


volumio:
Sorry, this can't be changed like that. I need this in place for some patches I need to apply later sometimes

macmpi:
I did keep it inside the "volumio" patch area.  `if [ "$PATCH" = "volumio" ]; then`
It is just out of the nested `if [ "$KERNEL_VERSION" = "4.4.9" ]; then` statement, so it should be equally fine with your patching mechanism.
No?

PR #199, opened 2017-06-07
Pi: bump next kernel test case to 4.9.31

macmpi:
As always, this change does NOT affect standard build of Volumio on 4.4.9


volumio:
I was just thinking the same! On this one we'll have the Allo digione support!

macmpi:
just need to add in `dacs.json` (along with Hifiberry Digi+ Pro which is supported by kernel too)

PR #200, opened 2017-06-09
Pi: apply standard Raspbian system configuration

macmpi:
Provided by [raspberrypi-sys-mods package](https://github.com/RPi-Distro/raspberrypi-sys-mods) used in all Raspian releases (**hence heavily tested with all Pi Models**).
In particular, this standard package most notably fixes:
- gpio_mem userland access (same 99-com.rules udev rule as previously - so no change actually)
- improve device-tree i2c driver load (i2c-modprobe rule) as per [this](https://www.raspberrypi.org/forums/viewtopic.php?p=1037943#p1037943)
- rfkill Bluetooth setting not persisting over reboot
- blacklist unmature rtl8192cu/rtl8xxxu wifi drivers (and conflicting with working 8192cu)
...

This PR also updates `cmdline.txt` [serial interfaces names consistently](https://www.raspberrypi.org/forums/viewtopic.php?f=29&t=138120) with used firmware.

**Notes:**
raspberrypi-sys-mods package is applied after `cmdline.txt` and `config.txt` are created, as it makes some consistency check on them. (they are not altered because their are already fine).
As post-install, we remove the only 2 unwanted changes (pi user sudoer, and ssh protection).

Tested on PiZero equipped with pHAT DAC.

### DIFF
--- a/scripts/raspberryconfig.sh
+++ b/scripts/raspberryconfig.sh
@@ -114,53 +114,9 @@ wget http://repo.volumio.org/Volumio2/wireless-firmwares/brcmfmac43143.bin -P /l
 echo "Installing WiringPi from Raspberrypi.org Repo"
 apt-get -y install wiringpi
 
-echo "adding gpio & spi group and permissions"
-groupadd -f --system gpio
-groupadd -f --system spi
-
-echo "adding volumio to gpio group and al"
-usermod -a -G gpio,i2c,spi,input volumio
-
 echo "Configuring boot splash"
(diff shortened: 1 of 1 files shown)


volumio:
Can you send the PR to volumionext branch?

macmpi:
Original PR has also value for main branch on 4.4.9: it really covers any release.
But I can duplicate if you want.

volumio:
Please send it to volumionext branch

PR #201, opened 2017-06-10
Make sure swapiness is always set correctly

macmpi:
Was not set at boot if swap file already existed.
now always set the same way


macmpi:
should this one go to volumionext too?

volumio:
yes please

PR #205, opened 2017-06-12
Update to latest LTS node-v6.11.0

macmpi:
For armv6 and x86.
Shouldn't we also install dedicated armv7 binaries for non-Pi armv7 boards ?

Please mind to:
- add the packages to the repo
- recompile the modules


volumio:
It has to be done in volumionext, don't worry I'll take care of it

PR #210, opened 2017-06-22
Refactor

PeterDaveHello:


volumio:
What are the advantages of such approach?

PeterDaveHello:
The origin indents were mixed with tabs and spaces, or maybe just be wrong, also, there were some deprecated, non-safe usages, the refactor is for better quality.

Some of the issues can be detected by [ShellCheck](https://github.com/koalaman/shellcheck/), maybe you'll be interested in the tool, even integrated with CI.

PR #216, opened 2017-07-01
Pi: superfluous apt-get update

macmpi:
just makes build longer...


volumio:
Correct. It would be great if you could apply the same to other platform files

PR #218, opened 2017-07-03
Allow /volumio_command_line_client/volumio.sh to be run NOPASSWD

xipmix:
This avoids having to do ```echo <password> | sudo -S ``` which causes issues if the volumio user password changes. See volumio/Volumio2#1192 and friends.

This PR is prepatory work for another one destined for Volumio2.


xipmix:
Dropping this - I've missed a couple of things.

PR #219, opened 2017-07-03
Removing superfluous apt-get update

macmpi:
same as https://github.com/volumio/Build/pull/216 for other devices as requested.

### DIFF
--- a/scripts/armbianconfig.sh
+++ b/scripts/armbianconfig.sh
@@ -62,7 +62,6 @@ fi
 rm /patch
 
 echo "Installing winbind here, since it freezes networking"
-apt-get update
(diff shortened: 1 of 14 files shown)


volumio:
Sorry, this is necessary because we do clean the apt cache sometimes in the patches

macmpi:
then you may want to also revert the Pi one accordingly #216

macmpi:
@volumio (this is old stuff, but just to make sure):
Since this PR has been cancelled and for consistency, don't you want to revert https://github.com/volumio/Build/pull/216 which originally accomplished same on Pi?

volumio:
I am checking now... thanks

PR #220, opened 2017-07-04
Disable powerdown mode for digital out.

leafy7382:
Digital output from all sound cards except the USB will enter a power down mode in a default of 5000 ms. TOSLINK/COAX being shutdown when idle causes issues with DACs that as relays in them such as Schiit Yggdrasil or NAD D1050. The relay triggers whenever the digital out turns on/off, causing excessive clicking noise and loss of first few seconds of music during the transition.

Adding this option will turn off the power down time and maintain digital out all the time.

See line 66 at linux/sound/soc/soc-core.c


volumio:
Clever, thanks

PR #222, opened 2017-07-04
give sudo permissions for two volumio_commandline_client scripts

xipmix:
volumio/Volumio2#1192 highlights the problems that arise when using ```echo <passwd> | sudo -S``` to get sudo access. Analysis of the code base finds 9 instances in ```commands/kernelsource.sh``` and 6 in ```volumio.sh```.  (Two other instances exist, which will not be addressed here).

In volumio.sh most instances are for calling ```systemctl``` which is already allowed under NOPASSWD. The remaining two instances are the scripts ```commands/kernelsource.sh``` and ```commands/pull.sh```. There seems little reason not to make these two developer scripts runnable with the sudo NOPASSWD: option. Doing so allows most of the other ```echo <passwd>``` instances to be removed also (separate pull request).

Tested on rpi2 with 2.201. The 'kernelsource' and 'pull' actions work as before.

### DIFF
--- a/scripts/volumioconfig.sh
+++ b/scripts/volumioconfig.sh
@@ -90,7 +90,7 @@ alias ifconfig="sudo /sbin/ifconfig"
 
 #Sudoers Nopasswd
 echo 'Adding Safe Sudoers NoPassw permissions'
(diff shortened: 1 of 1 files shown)


xipmix:
rebased after months of bitrot and got into a pickle. resubmitting shortly.

PR #232, opened 2017-07-12
commenting out a fail init script

gkkpch:


gkkpch:
Refers to the config build script for Voltastream (WIP)

PR #233, opened 2017-07-13
Support for Kernel Updater on X86

gkkpch:


volumio:
Thanks so much Gè... I'll build and ask the community to test it here:
https://volumio.org/forum/help-needed-test-the-new-updater-t7057.html

gkkpch:
Good, I have not distributed the 4.9 link yet as I was "thrown" out of my room today because of another viewing :D
Hope to have an answer to this soon, like the viewings but not what it means for my workroom ;)
Cheers - Gé

PR #236, opened 2017-07-23
x86 kernel updater test prep

gkkpch:


scripts/x86image.sh
```diff
@@ -82,11 +82,16 @@ sudo mount -t vfat "${BOOT_PART}" /mnt/volumio/rootfs/boot
 cp scripts/x86config.sh /mnt/volumio/rootfs
 if [ ! -d platform-x86 ]; then
   echo "Platform files (packages) not available yet, getting them from the repo"
-  git clone http://github.com/volumio/platform-x86
+  git clone http://github.com/gkkpch/platform-x86
```
volumio: We should clone from volumio's repo

PR #247, opened 2017-08-05
Avoid SMB1 protocol

xipmix:
SMB1 is dangerous (e.g. WannaCry malware) and while still the default is best avoided.
smb2_02 is supported by Windows Vista and later, which should be a resonable choice.
smb2_10 (Win7 and later) might be better, what do people think?

This was requested in volumio/Volumio2#878. Not tested yet as I rarely use samba
and have never used the shares exposed by volumio.
Anyone who wants to jump in and test or comment, please do.

### DIFF
--- a/volumio/etc/samba/smb.conf
+++ b/volumio/etc/samba/smb.conf
@@ -10,6 +10,7 @@ local master = no
 preferred master = no
 os level = 30
 client min protocol = smb2_02
+server min protocol = smb2_02
 
 [Internal Storage]
         comment = Volumio Internal Music Folder
--- a/volumio/etc/samba/smb.conf
+++ b/volumio/etc/samba/smb.conf
@@ -9,6 +9,7 @@ wins support = yes
 local master = no
 preferred master = no
 os level = 30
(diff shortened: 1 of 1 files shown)


gerritv:
I edited the smb.conf file per the new additions, rebooted. Adding a new drive for Windows Server 2012R2 failed with Host is down. Adding vers=2.0 in 'options' and the share was registered.

Volumio 2.246

gerritv:
Upon further reflection and more reading, the problem is with the mount cifs command contents, not with Samba configuration. Without vers=2.0 on the mount command line the operation will fail to mount the share.

xipmix:
I'm working on a PR for that side of things but that I am more able to actually test - so I'm doing that before submitting. Here I just wanted to ensure that volumio is doing its part to avoid smb1.

volumio:
We should take those decisions carefully: this prevents older windows or clients to access volumio shared folder. What risks do we have in leaving default v1?

volumio:
See gerritv comment, too much of a breaking change

PR #252, opened 2017-08-12
Pi: move to Kernel 4.9.41

macmpi:
Several improvements, including better support for several i2s DACs (allo, etc).
Has been also pushed as new "stable" reference branch

### DIFF
--- a/scripts/raspberryconfig.sh
+++ b/scripts/raspberryconfig.sh
@@ -56,7 +56,7 @@ sudo curl -L --output /usr/bin/rpi-update https://raw.githubusercontent.com/Hexx
 touch /boot/start.elf
 mkdir /lib/modules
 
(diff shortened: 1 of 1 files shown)


volumio:
Thanks!

macmpi:
You're welcome.
Now, I imagine kernel will never go back to 4.4.9 anymore: so we could probably do some cleanup in `raspberryconfig.sh`: remove `case` statement and 4.4.9 specific drivers installs.

Also I used to adapt [`cpuinfo` device detection ](https://github.com/volumio/Volumio2/pull/1127/files)so that it could work with both kernel families during transition tests: now we could probably nail it down to only `BCM2835` that all Pi models returns since 4.8.y.

Would you agree we clean that up now 4.9.y migration is done ?

PR #256, opened 2017-09-03
Replace hardcoded pieces of the initrd script with variables so that …

DecayFi:
…the boot device can be changed completely. To make the system boot from /dev/sda*, "bootdev=sda dps= genpnames" needs to be added to the kernel cmd line. dps is the device partition seperator. This was tested on both booting from sdcard and usb stick. The data partition resize is working as well, however the USB update will be disabled if the system boots from usb to avoid that the updater kills the boot medium

### DIFF
--- a/scripts/initramfs/init
+++ b/scripts/initramfs/init
@@ -19,6 +19,12 @@ mdev -s
 USE_KMSG="yes"
 HWDEVICE="empty"
 BOOTDEV="mmcblk0"
+#Device/Partition Separator i.e. the character between parentdev and partition index
+DPS="p"
+#default values:
+IMGPART="/dev/${BOOTDEV}${DPS}2"
+DATAPART="/dev/${BOOTDEV}${DPS}3"
+BOOTPART="/dev/${BOOTDEV}${DPS}1"
 
 # Display a message or print directly to /dev/kmsg
 print_msg() {
@@ -68,6 +74,20 @@ do
     ;;
   bootdev)
     BOOTDEV=$value
+    ;;
+  dps)
+    DPS=$value
+    ;;
+  genpnames)
+    IMGPART="/dev/${BOOTDEV}${DPS}2"
+    DATAPART="/dev/${BOOTDEV}${DPS}3"
+    BOOTPART="/dev/${BOOTDEV}${DPS}1"
+    ;;
+  bootpart)
+    BOOTPART=`parse_disk $value`
+    ;;
+  datapart)
+    DATAPART=`parse_disk $value`
     ;;
   hwdevice)
     HWDEVICE=$value
@@ -122,6 +142,9 @@ fi
 
 print_msg IMGPART=${IMGPART}
 print_msg IMGFILE=${IMGFILE}
+print_msg DATAPART=${DATAPART}
+print_msg BOOTPART=${BOOTPART}
+
 if [ ! -z "${BOOTDELAY}" ]; then
   print_msg "Boot delay (except first time) will be ${BOOTDELAY} seconds"
 fi
@@ -161,7 +184,7 @@ if [ ! -e "/mnt/imgpart/volumio_factory.sqsh" ]; then
   print_msg "Factory image created"
   print_msg "Creating archive for factory kernel..."
   mkdir /mnt/factory
-  mount -t vfat /dev/${BOOTDEV}p1 /mnt/factory
+  mount -t vfat ${BOOTPART} /mnt/factory
   tar cf /mnt/imgpart/kernel_current.tar -C /mnt/factory .
   cp /mnt/imgpart/kernel_current.tar /mnt/imgpart/kernel_factory.tar
   umount /mnt/factory
@@ -171,16 +194,16 @@ elif [ ! -z "${BOOTDELAY}" ]; then
   sleep ${BOOTDELAY}
 fi
 
-print_msg "Checking for USB updates"
+print_msg "Checking for USB updates if you did not boot from USB..."
 [ -e /dev/sda1 ] || mdev -s
-if [ -e /dev/sda1 ]; then
+if [ -e /dev/sda1 ] && [ ! "/dev/sda1" ==  "${BOOTPART}" ]; then
   [ -d /mnt/usb ] || mkdir /mnt/usb
   mount -t auto /dev/sda1 /mnt/usb
   #If there is a firmware file inside the usb
   if [ -e /mnt/usb/*.fir ]; then
     print_msg "Firmware found, updating will take a few minutes, please wait..."
     mkdir /mnt/boot
-    mount -t auto /dev/${BOOTDEV}p1 /mnt/boot
+    mount -t auto ${BOOTPART} /mnt/boot
     #when the partitions are mounted we can launch the update script
     volumio-init-updater
     sync
@@ -193,7 +216,7 @@ if [ -e /dev/sda1 ]; then
   if [ -e /mnt/usb/factory_reset ]; then
     print_msg "Factory Reset on USB"
     mkdir /mnt/factory
-    mount -t auto /dev/${BOOTDEV}p1 /mnt/factory
+    mount -t auto ${BOOTPART} /mnt/factory
     echo " " > /mnt/factory/factory_reset
     umount /mnt/factory
     rm -r /mnt/factory
@@ -202,7 +225,11 @@ if [ -e /dev/sda1 ]; then
   umount /dev/sda1
   rm -r /mnt/usb
 else
-  print_msg "No USB device detected (when incorrect, try adding 'bootdelay=5' to your boot cmdline)"
+  if [ "/dev/sda1" == "${BOOTPART}" ]; then
+    print_msg "Not checking for firmware if you boot from USB. Sorry!"
+  else 
+    print_msg "No USB device detected (when incorrect, try adding 'bootdelay=5' to your boot cmdline)"
+  fi
 fi
 
 
@@ -225,10 +252,10 @@ VOLUMIO_VERSION="$(cat /mnt/static/etc/os-release | grep VOLUMIO_VERSION)"
 #if there is factory file then format data partition
 #
 mkdir /mnt/factory
-mount -t auto /dev/${BOOTDEV}p1 /mnt/factory
+mount -t auto ${BOOTPART} /mnt/factory
 if [ -e "/mnt/factory/factory_reset" ]; then
   print_msg "Executing factory reset"
-  mkfs.ext4 -F -E stride=2,stripe-width=1024 -b 4096 /dev/${BOOTDEV}p3 -L volumio_data
+  mkfs.ext4 -F -E stride=2,stripe-width=1024 -b 4096 ${DATAPART} -L volumio_data
   print_msg "Factory reset executed: part I - User DATA Part"
   tar xf /mnt/imgpart/kernel_factory.tar -C /mnt/factory
   print_msg "Factory reset executed: part II - Kernel"
@@ -245,7 +272,7 @@ if [ -e "/mnt/factory/factory_reset" ]; then
 fi
 if [ -e "/mnt/factory/user_data" ]; then
   print_msg "Deleting User Data"
-  mkfs.ext4 -F -E stride=2,stripe-width=1024 -b 4096 /dev/${BOOTDEV}p3 -L volumio_data
+  mkfs.ext4 -F -E stride=2,stripe-width=1024 -b 4096 ${DATAPART} -L volumio_data
   rm /mnt/factory/user_data
   print_msg "User Data successfully deleted "
 fi
@@ -255,7 +282,7 @@ rm -r /mnt/factory
 
 # if the update failed before completion
 mkdir boot
-mount -t vfat /dev/${BOOTDEV}p1 /boot
+mount -t vfat ${BOOTPART} /boot
 if [ -e "/boot/update_process" ]; then
 print_msg "Previous update attempt failed, restoring fallbacks"
   cp /mnt/imgpart/kernel_fallback.tar /mnt/imgpart/kernel_current.tar
@@ -290,8 +317,8 @@ if [ -e "/boot/resize-volumio-datapart" ]; then
 print_msg "Re-sizing Volumio data partition"
   END="$(parted -s /dev/${BOOTDEV} unit MB print free | grep Free | tail -1 | awk '{print $2}' | grep -o '[0-9]\+')"
   parted -s /dev/${BOOTDEV} resizepart 3 ${END}
-  e2fsck -fy /dev/${BOOTDEV}p3
-  resize2fs /dev/${BOOTDEV}p3
+  e2fsck -fy ${DATAPART}
+  resize2fs ${DATAPART}
   print_msg "Volumio data partition succesfully resized"
   parted -s /dev/${BOOTDEV} unit MB print
   rm /boot/resize-volumio-datapart
@@ -304,7 +331,7 @@ rm -rf /boot
 # 4) mount a filesystem for write access to the static image
 # unclear: memory size? -o size=1024M
 [ -d /mnt/ext ] || mkdir -m 777 /mnt/ext
-mount -t ext4 -o noatime /dev/${BOOTDEV}p3 /mnt/ext
+mount -t ext4 -o noatime ${DATAPART} /mnt/ext
 
 [ -d /mnt/ext/dyn ] || mkdir -m 777 /mnt/ext/dyn
 [ -d /mnt/ext/union ] || mkdir -m 777 /mnt/ext/union
@@ -325,6 +352,14 @@ mount --move /mnt/static /mnt/ext/union/static
 mount --move /mnt/imgpart /mnt/ext/union/imgpart
 
 chmod -R 777 /mnt/ext/union/imgpart
+
+#idea: Mount /boot already here to that systemd does not have to mount it an we
+# can be flexible. However this should be confirmed by somebody who is more familiar 
+# with systemd and who can tell what fstab is used for in the system
+
+[ -d /mnt/ext/union/boot ] || mkdir -m 777 /mnt/ext/union/boot
+mount -t vfat ${BOOTPART} /mnt/ext/union/boot
+
 
 umount /proc
 umount /sys
--- a/scripts/initramfs/init
+++ b/scripts/initramfs/init
@@ -324,7 +324,8 @@ print_msg "Re-sizing Volumio data partition"
   END="$(parted -s /dev/${BOOTDEV} unit MB print free | grep Free | tail -1 | awk '{print $2}' | grep -o '[0-9]\+')"
   parted -s /dev/${BOOTDEV} resizepart 3 ${END}
   e2fsck -fy ${DATAPART}
-  resize2fs ${DATAPART}
+  #force resize as we did just run e2fsck. Resize2fs seems ro not detect this occasionally
+  resize2fs -f ${DATAPART}
   print_msg "Volumio data partition succesfully resized"
   parted -s /dev/${BOOTDEV} unit MB print
   rm /boot/resize-volumio-datapart
--- a/scripts/initramfs/init
+++ b/scripts/initramfs/init
@@ -55,6 +55,8 @@ parse_disk() {
   fi
 }
 
+DO_GEN=no
+
 for p in ${CMDLINE};
 do
   key=${p%%=*}
@@ -79,9 +81,7 @@ do
     DPS=$value
     ;;
   genpnames)
-    IMGPART="/dev/${BOOTDEV}${DPS}2"
-    DATAPART="/dev/${BOOTDEV}${DPS}3"
-    BOOTPART="/dev/${BOOTDEV}${DPS}1"
+    DO_GEN=yes
     ;;
   bootpart)
     BOOTPART=`parse_disk $value`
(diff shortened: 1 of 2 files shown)


DecayFi:
Regarding uuid, I was not sure how you can get the parent drive of a partition you only have the UUID of, which seems to be necessary for the resizing part when using parted. And I only have the possibility to test on Pi3 and Pi Zero W, so what I commited seems to run fine when booting Raspberry Pi from USB but still using the classical device file approach. Are you planning to switch the mainline to uuid anytime soon?

gkkpch:
yes, I will finish testing it, especially wrt BOOTDEV, UUID and Rock64. There will be nothing stopping it going into the main repo.

DecayFi:
So is there more to do on this topic or is everything adressed? Did you already test the resizing using uuids?

gkkpch:
Won't be able to continue until coming Thursday (28.09), I'll post my results here.


On 24 Sep 2017, at 18:58, DecayFi <notifications@github.com> wrote:

So is there more to do on this topic or is everything adressed? Did you already test the resizing using uuids?

macmpi (commented):
can't we keep /boot mount in each <platform>config.sh fstab ?

DecayFi:
The problem is that fstab is hardcoded at compile time and can not easily be chaned at run time. Maybe this could remain if a uuid was used. Us there an usecase that requires fstab?

If there is a mount in initrd and one in fstab, currently the entry in fstab "wins", so the fstab entry will be the mount visible to volumio. So Volumio will not be able to change the config.txt  and hence not to change dacs or plugins.

macmpi:
Once the squash image is mounted, can't the script modify `/etc/fstab` in the specific way needed in case of USB boot, and then `chroot`?
Such modified fstab will survive across reboots in the overlayfs.
Would that make sense?

gkkpch:
moving the /boot mount to initrd has the advantage we only need to change the mount in one place. Especially with using UUID for newer devices, it would cause extra complexity when using the kernel updater. It can be done, see x86, but I was planning to change that one along this solution as it is cleaner (I think).

DecayFi:
The squashfs is a read-only filesystem afaik, so it can not be modified at runtime.  So one would have to make a file or link in the overlay fs to mask the hardcoded fstab at every boot. Is there an advantage to make fstab do the mount?

macmpi:
Once squash is mounted in overlayfs, files (including `fstab`) can be modified at runtime: it's the purpose of overlayfs. So I guess init script can technically tweak baseline `/etc/fstab` at runtime before chroot to manage such exception in this particular usecase/platform.

I guess current design tries to keep all runtime-related stuff (including all core systems mounts) within the same part of the images layout: I guess it's probably easier to maintain across all supported platforms: nothing scattered into different places.
I thought it would make sense such boot exceptions tweaks are done, if possible, in a way that does not create disruption in that design. However @gkkpch and @volumio are the real architects here, and my interpretation may be wrong.

I indeed have an interest in keeping this design as-is (keep init minimal to just do updates and bare mounts for chroot): squash images are then more self-contained and therefore easily transportable to other environments (such as berryboot): i.e. mounting `/boot` in init would break such seamless transport capability.
I know this is not supported and corner-case use-case, so I will adapt of course. However if this benefit can survive without too much pain, that's nice!

DecayFi:
Never tried berryboot. However I don't think that they like hardcoded partition names. Is there a link where can see what needs to be adjusted for berryboot?

gkkpch:
@DecayFi,
Changing fstab is easy after mounting the overlay, but I would still like to do the way you suggested. It offers more flexibility and is easier to maintain, also for newer devices with additional boot options. 
Unless there is a reason why we would need to stick to fstab. I can't think of one yet...

DecayFi:
@macmpi: Couldn't you just create a berryboot-init script to make the mount? I also read your post in the forum about setting up volumio in berryboot. If you do not modify the squashfs in the volumio image, does this mean you mount the berryboot /boot partition under volumio as well? Or did you find another way to mount the /boot partition as well?
@gkkpch : Even with the latest submit it is still hardcoded that the data partition must be partition number 3 of BOOTDEV. Does it make sense to extract the parent and the parition number out of the data-partition as well? This would allow usecases where the datapart is on a different device than the bootdev and still do resizing. However, I am not sure how far the flexibility should go.  This should be possible, but would need lsblk and libsmartcols.so.1 to be part of the initrd.

macmpi:
@DecayFi : for curiosity you can check their [initrd](https://github.com/macmpi/berryboot/blob/berryboot2015/buildroot-2015.02/package/berrybootgui2/init), and see how they manage startup with client OS squash images. Interestingly, this bootloader does support USB boot. Mindset is do bare minimum and let client OS do their own mounts (like `/boot`) should they need so (and don't be scared by aufs stuff, let's pretend it's similar to overlayfs in the context of this discussion).

@gkkpch : I understand your choice. However in that case, do you imply all of fstab mounts will be moved to initrd for all platforms? Or will that be only for `/boot ` and/or depend on platforms?

macmpi:
@DecayFi our messages crossed. Currently it just works nicely out-of-the-box as Volumio squash does its own mounts for what it needs: that's the beauty of the "self-contained" design.
If Volumio layout changes, I'll try to figure-out a way. I do not find berryboot-init script satisfactory as modifying the squash image is a big burden. I proposed a PR to make those scripts more dynamic (for plenty of other benefits) but maintainer does not seem inclined to accept. I'll find another way around probably. However, if looking at their code helps your design work, that's great.

gkkpch:
I see your point and was thinking of it, including the use of a BOOTPART UUID alternative to find the BOOTDEV. Can I answer coming Thursday? I'm 2000km away from home turf and only have access to simple mobile just now. 

On 25 Sep 2017, at 22:03, DecayFi <notifications@github.com> wrote:

@macmpi: Couldn't you just create a berryboot-init script to make the mount? I also read your post in the forum about setting up volumio in berryboot. If you do not modify the squashfs in the volumio image, does this mean you mount the berryboot /boot partition under volumio as well? Or did you find another way to mount the /boot partition as well?
@gkkpch : Even with the latest submit it is still hardcoded that the data partition must be partition number 3 of BOOTDEV. Does it make sense to extract the parent and the parition number out of the data-partition as well? This would allow usecases where the datapart is on a different device than the bootdev and still do resizing. However, I am not sure how far the flexibility should go. This should be possible, but would need lsblk and libsmartcols.so.1 to be part of the initrd.

gkkpch:
@macmpi/ @DecayFi, just a quick answer: no, the idea is only to remove /boot as the device it is mounted on cannot be defined at build time in all cases.
/etc/fstab itself is not under discussion and should remain standard for all platforms. 
I'm risking to repeat myself here, but there are platforms other than PI where we cannot hardcode /dev/<something> anymore for image and boot partition at build time because we simply cannot know what <something> is going to be. X86 was the first one, new arm devices now seem to follow. We need to add flexibility and @DecayFi's approach is a very good step in a new direction. I know, BOOTDEV may not be definite, perhaps I'll add a BOOTPART option with UUID to locate BOOTDEV. But most of it is now there already and seems to work.



On 25 Sep 2017, at 22:05, macmpi <notifications@github.com> wrote:

@DecayFi : for curiosity you can check their initrd, and see how they manage startup with client OS squash images. Interestingly, this bootloader does support USB boot. Mindset is do bare minimum and let client OS do their own mounts (like /boot) should they need so (and don't be scared by aufs stuff, let's pretend it's similar to overlayfs in the context of this discussion).

@gkkpch : I understand your choice. However in that case, do you imply all of fstab mounts will be moved to initrd for all platforms? Or will that be only for /boot and/or depend on platforms?

macmpi:
One (silly) idea maybe.
Since it's only about `/boot` and that this mount is essentially needed for the purpose of Volumio2 app (tweaking DACs for instance), what about this mount happens within some of Volumio2 app early startup initialization, like [here](https://github.com/volumio/Volumio2/blob/master/app/plugins/system_controller/system/index.js#L27) for instance?
I guess at that runtime we can search/find/mount the proper device in a programmatic (not hardcoded) way, no?
With that, things will be kept "self-contained" (squash image does the things it needs for itself), valid for all configurations/platforms.
Makes sense?

gkkpch:
this would not work as we can't get to it in Initrd when we need to update the UUID in case of kernel updates

macmpi:
sure, but I was talking just about "normal operations".

So to explain my idea in more details:
Principle remains as it is currently: let initrd just take care of bootloading and special "meta-operations" (updates, reset to factory, wipe data,...): it does all it needs for its own operations, and only that.
Anything that is Volumio runtime & app specific pertains to squash and Volumio2 app: should they need to mount stuff, they do it on their own.

With that in mind, initrd would mount `/boot` only when it needs to (i.e. for kernel updates), and eventually unmount (though in kernel update case, system does reboot right-after so probably no need to unmount). If there is no kernel update, initrd does not mount `/boot`.
On the other side, proposal is to let Volumio2 app mount `/boot` as it needs it, for its own stuff (config management).

Don't you think such design separation can offer satisfactory and clean compromise?
Since there is a clear logic boundaries, its not difficult to maintain either.

DecayFi:
@macmpi: Do you have a good idea where volumio should take the information which partition to mount as /boot from? Probably one could parse the kernel cmdline again, but does berryboot allow you to start the squashfs with a custom kernel commandline? Do you see a solution that works for both usecases? 

On the other hand, isn't this scenario exactly what the berryboot-init scripts are made for?

macmpi:
In Volumio's `initrd`, finding the unmounted `/boot` partition in a programmatic way could be achieved [somehow like this](https://github.com/maxnet/berryboot/blob/berryboot2015/BerrybootGUI2.0/installer.cpp#L139), searching for `volumio.initrd` file as a clue in `findBootPart()` for instance.
Such search routine could possibly be within a script which could be called either from initrd or Volumio2 App, so that each one can find what it needs when it needs it.

gkkpch:
This whole discussion is obviously about berryboot, which is not running on a platform other than a Raspberry PI (volumio is not supported on Banana and Orange). 
There are no other reasons?
So the real question is, how can we implement it and still have berryboot support?
Can we implement the changes and make an exception not to mount /boot in initrd when berryboot.img is present and assume /boot is mounted somewhere else in that case but not via /etc/fstab?




On 27 Sep 2017, at 07:15, macmpi <notifications@github.com> wrote:

In Volumio's initrd, finding the unmounted /boot partition in a programmatic way could be achieved somehow like this, searching for volumio.initrd file as a clue in findBootPart() for instance.
Such search routine could possibly be within a script which could be called either from initrd or Volumio2 App, so that each one can find what it needs when it needs it.

macmpi:
I do not intend to make this Berryboot discussion: it serves as an example.
I think the overall idea is: can we keep core Volumio app (along overall squash) as independent as possible from initrd (except for "meta stuff" like updates & al of course)?

It was the case so far, but, with the upcoming dynamic partition improvements, can we do this PR in a way **that keeps such independence**?
Value of such independence is to allow use of different bootloading: let's treat Berryboot as "just" a testcase. (I would advocate nothing in Volumio should be tied to Berryboot specifics).

The proposed PR improvements introduces **only one issue** with this independence: it's the way it moves `/boot` mount from `fstab` to `initrd`. If we can keep that mount back to system/app side (the part that uses it), then we maintain current bootloader independence.

Dynamic `/boot` partition detection brings value to overall dynamic partition improvements I think, and it should NOT linked to Berryboot specifics or anything else: I just proposed a sample code, and if there are better methods to do it, that's equally fine.

gkkpch:
Well, in that case you lost me completely. I have not seen one reason yet for Volumio why we need to keep /boot mount outside initrd. I would like turn it around and let /boot mount in initrd be the standard, it makes boot detection easy, flexible and keeps everything related to the boot process and configuration issues separated from the application side. 

On 27 Sep 2017, at 09:05, macmpi <notifications@github.com> wrote:

I do not intend to make this Berryboot discussion: it serves as an example.
I think the overall idea is: can we keep core Volumio app (along overall squash) as independent as possible from initrd (except for "meta stuff" like updates & al of course)?

It was the case so far, but, with the upcoming dynamic partition improvements, can we do this PR in a way that keeps such independence?
Value of such independence is to allow use of different bootloading: let's treat Berryboot as "just" a testcase. (I would advocate nothing in Volumio should be tied to Berryboot specifics).

The proposed PR improvements introduces only one issue with this independence: it's the way it moves /boot mount from fstab to initrd. If we can keep that mount back to system/app side (the part that uses it), then we maintain current bootloader independence.

Dynamic /boot partition detection brings value to overall dynamic partition improvements I think, and it should NOT linked to Berryboot specifics or anything else: I just proposed a sample code, and if there are better methods to do it, that's equally fine.

macmpi:
New Volumio2 design with squashfs made it really easy "transport volumio" on a different bootloader(s), which allows to run the app in different ways (dynamically test, compare, debug different Volumio squash images without reflashing, multiple distributions on same partition, run from networked iSCSI partitions, etc).
Granted it was probably not the purpose of original squashfs design choice, but the way it turned-out was very favorable from that viewpoint.
Other popular distributions (LibreElec, ...) are also somehow designed in the same way.

It is indeed non-standard (advanced) use of Volumio, but that newly-acquired independence between bootloader & core app/system was great value so far.
The more initrd integrates specific runtime dependencies within itself, the more this versatility gets compromised: I did not expect mounts intended for application-side would have to migrate into initrd at some point...

I totally understand if maintaining such independence may not be a design priority, or is felt too much of cost in terms of complexity sometimes: I'll find independent hack-ish workarounds for my use-cases.

Thanks for the discussion: I hope it helped raising awareness to try to keep such sort of initrd changes as minimal as possible, whenever possible.

DecayFi:
@macmpi: The original idea was to remove hardcoding from the squashfs, as there seem to be platforms where something other than mmcblk0p2 has to be mounted as /boot and like this the same squashfs could ne shared among many platforms. If there is no hardcoding in the squashfs, there needs to be a source for this information. Your proposal to detect the at runtime seems to point also to the kernel cmdline (if I understood this correctly when having a short glance). So even if one would create something in volumio startup to do the mount (which I am unsure how much effort this would be), would it still satisfy your berryboot usecase? Can you add all options you need for all os Images to your berryboot kernel cmdline? Or is there a way in berryboot to have an individual kernel cmdline per os image?

Also, with an appended berryboot-init the different images should still work. The appending could even be done in your install-script in the forum...

macmpi:
@DecayFi in fact the alternative I was getting at is to somehow translate the code below (or anything better) as a bash script or native node.js and invoke it within Volumio app to mount `/boot`, intead of mounting within initrd bootloader.
But I understand @gkkpch 's preference to stick with initrd mount, as it is simpler.

I'll workout the specific berryboot setup tweaks separately.

```
QString Installer::findBootPart()
{
    /* Search for boot partition containing volumio.initrd (or other better clue) */
    QString part;
    QString dirname  = "/sys/class/block";
    QDir    dir(dirname);
    QStringList list = dir.entryList(QDir::Dirs | QDir::NoDotAndDotDot);

    foreach (QString devname, list)
    {
        /* Only search first partition and partitionless devices. Skip virtual devices (such as ramdisk) */
        if ((devname.right(1).at(0).isDigit() && !devname.endsWith("1"))
                || QFile::symLinkTarget("/sys/class/block/"+devname).contains("/devices/virtual/"))
            continue;

        if (QProcess::execute("mount -t vfat -o ro /dev/"+devname+" /boot") == 0)
        {
            if (QFile::exists("/boot/volumio.initrd"))
            {
                qDebug() << "Found volumio.initrd at" << devname;
                part = devname;
            }

            QProcess::execute("umount /boot");
        }

        if (!part.isEmpty())
            break;
    }

    return part;
}
```

DecayFi:
Wow, what a workaround... But in my opinion it's still not really clean, as it does not evauluate the chioce the user made with the kernel cmdline. From looking at the code I would not take for granted that this will not fail, not take very long or even detect the wrong boot if one happens to have a usb sick with a volumio.initrd inserted. Also, is it somehow ensured that this code terminates on every platform with every configuration?

On the other hand, shouldn't something like
```
sudo apt-get install squashfs-tools
echo \#\!/bin/\sh >berryboot-init
echo mount -t vfat -o defaults,utf8,user,rw,umask=111,dmask=000 \/dev\/mmcblk0p1 \/squashfs\/boot >>berryboot-init
chmod a+x ./berryboot-init
mkqsuashfs ./berryboot-init volumio_current.sqsh 
rm ./berryboot-init

```
in your install / update script do the same thing for you? Even the berryboot os porting guide recommends to remove hardcoded mount entries from fstab in any OS to be run under berryboot...

macmpi:
@DecayFi I agree this may not be bulletproof in current form (did a quick sh translation).

As a fallback on the BB install scripts, I'll probably either modify a copy of new `/etc/fstab` or take the pill of modifying images.
Thanks for the discussion guys.

gkkpch:
We are faced with newer arm devices, but using /boot with a hard coded device in fstab (or mounting it anyware else in the application part) is an absolute no-go. This is not really new, we already have  experience from x86, having to search for the boot device. 
Having a UUID in fstab appears to be messy, it makes OTA updating unnecessary complicated.

As the discussion showed, it appears this issue is too important just to dismiss it with taking /boot out of /etc/fstab.
Therefore I will implement the @DecayFi changes (adding to it), but leave /etc/fstab and /boot mounting for arm devices as is at the moment. 

For the platform that currently really needs removing /boot from fstab (Rock64) in order to keep control over changing boot device UUIDs in initramfs, I will make an exception in initrd. 
The build scripts other than the one for Rock64 will remain unchanged.  
Let's gather experience with this and see what other options we can get come up with... 

btw. @DecayFi 's changes can also be applied partially to x86, hopefully get this done this weekend as a start.

DecayFi:
Not sure whether I understand this correctly, but are you saying that the plan is to not apply these changes for ARM platforms and leave fstab hardcoded to sdcard?

gkkpch:
No, I was going to implement the changes but leave /boot mount in fstab for the moment and make an exception for rock64 in initrd and exclude /boot from fstab in that particular case. 
But another option would be to always have 2 fstabs, one with and one without mounting /boot.
When using fstab with /boot, the boot device is limited to it being hardcoded. 
Without /boot in fstab, mounting /boot will take place in initrd. To activate this option, we would need a new boot parameter 'mountbootinitrd'. 
Omitting it presumes default with fstab.
I know, this is not ideal either, but it does keep the current functionality while allowing the other option through a commandline switch.





On 29 Sep 2017, at 08:52, DecayFi <notifications@github.com> wrote:

Not sure whether I understand this correctly, but are you saying that the plan is to not apply these changes for ARM platforms and leave fstab hardcoded to sdcard?

macmpi:
@gkkpch Thanks for your kind efforts and understanding to try mitigating the issue I raised, but it is clear accommodating it may make things unnecessarily complicated.

I finally determined I'll make a workaround on my side to manage such kind of different bootloader situation: so if my comments were the only gating reason, feel free to mount within `initrd` for all platforms, as originally thought.
Thanks again to both of you.

macmpi:
Just a few more thoughts on this.
One key concern with keeping `/boot` mounted at all times is data corruption risks particularly when using SD card for `/boot` (Achilles' heed on Pi), while in fact Volumio2 app-side does not actually use it very often (just for exceptional DAC changes, and software updates flags if I am correct - any other?).

So, it would be nice if Volumio2 app-side was more frugal about `/boot` use, and does `mount`/`umount` operations on the very seldom cases it actually needs `/boot`.
Agreed, for the app to do that, it would need to (very simply and reliably) know which device corresponds to `/boot`, since we want to be flexible (could be SD, some USB,...).

Actually there's an easy solution: indeed do not put `/boot`-related line into Built-in image `/etc/fstab`,  BUT rather have `init` script **inject that info in an overlayed copy of `/etc/fstab` at each boot**, together with `noauto` option.
Then App-side can find relevant info at runtime to do `/boot` mount when needed:
device path: `cat /etc/fstab | egrep "/boot" | awk '{print $1}'`
(options eventually: `cat /etc/fstab | egrep "/boot" | awk '{print $4}'`)

What do you think?

PS: BTW, while DAC settings have to be on `/boot`, how-about having the update flags in the overlayfs (or in `/imgpart`), hence reducing `/boot` mounts to extreme minimum (barely once in life).

macmpi:
Or, maybe simpler in this context, just mount `/boot` read-only by default, and have Volumio2 app do `mount -o remount,rw /boot` before any write and `mount -o remount,ro /boot` after.
Should be enough to prevent corruption on FAT filesystem (may not fully apply to ext3/ext4).

matthuisman:
Thanks

FYI for who wrote that script - I don't think you need the partition separator code.

It's this logic:
If the device ends with a digit (mmcblk0)- you add the 'p'
If the device ends with a letter (sda) - you don't

macmpi:
@matthuisman thanks for reviving this.
I think you may want to base your work from newer [init.nextarm](https://github.com/volumio/Build/blob/master/scripts/initramfs/init.nextarm) file as per @gkkpch suggestion, and submit PR (even as work in progress) as it is much easier to compare.

I now tend to agree we could remove `/boot` from `fstab`, but still think we need to consider potential data corruption on this `vfat` partition seriously.
Therefore we might not have it always mounted RW, but rather mount it RW temporarily in rare cases when needed (this means making limited changes in volumio app side too).
(BTW we could limit read/writes to it by moving update flag files elsewhere too, only doing `config.txt` alterations)

@gkkpch what do you think?

matthuisman:
@macmpi 

Couldn't all the writes to /boot from the app instead use the image or data partitions?
It's just the simple "flag files" that need to be written to /boot?

Why couldn't these be written to the image or data partitions?
Then after init, /boot only needs to be r/o 
Or.... it wouldn't even needed to be mounted in the app at all.

I thought it would make more sense to put the format needed flags etc in the /image partition as this is mounted first anyway.

Leaving the /boot partition alone as much as possible seems like a good idea to me.

It could just be mounted r/w in the init when a kernel update is applied.

macmpi:
DAC selection has to be written in `/boot/config.txt` since overlays need to be loaded at boot time.
Some Plugins might also need to alter (at install or in setup) `/boot/config.txt` for same reason.
I would think those are the only required writes (assuming "flags"-things can be done elsewhere).

So maybe we could have a javascript API to be called for writing to `/boot/config.txt`: that API would do the remount RW, do write & remount RO?
@volumio  what do you think?

ashthespy:
Closing this for now -- it will get tackled with the move to the new modular `init` (https://github.com/volumio/Build/pull/515)

PR #261, opened 2017-09-17
Pi: Add TauDAC-DM101 Firmware

taudac:


macmpi:
@taudac : 
Any plan to submit your driver & overlay for inclusion in raspberry kernel releases (like other DAC makers do)?
Or do you plan to make some automation to release & make available your package at each time kernel/firmware is released [here](https://github.com/Hexxeh/rpi-firmware/)?

This would greatly help as we sometimes bump Pi kernel versions in this build, and this might break taudac package compatibility if not available in time...

PR #264, opened 2017-09-29
Clean up after building initramfs

### LINKED ISSUE #253: Please clean /var/tmp/ after build
I mount the volumio_current.sqsh in volumio-2.246-2017-07-31-pi.img.zip to inspect its content. 
`/var/tmp` is not clean and occupied by several mega bytes:

> $ ls var/tmp
mkinitramfs_B4V371  mkinitramfs_IbKY5z
$ sudo du -hs var/tmp
292M    var/tmp

Also, I suggest to clean up unused file, like man and doc files using dpkg filters:
 * https://wiki.debian.org/ReduceDebian
 * https://wiki.ubuntu.com/ReducingDiskFootprint

xipmix:
To ensure we always run the exit trap, no matter how we exit, trap signal 0.
Tested for arm build.

Fixes #253


volumio:
Tried that, but the build size does not decrease by a single bit...

xipmix:
So without the patch there are no subdirectories in /var/tmp either ? Or they are there and not removed? When I tested the patch, /var/tmp in the squashfs was clean.

PR #266, opened 2017-10-05
Pi: move kernel to 4.9.51

macmpi:
Has been flagged as "stable"  about 2 weeks ago: many fixes since .41 including few related to sound issues.
Was holding submitting it, hoping longstanding Pi3/PiZW wireless fix would emerge soon, but we should probably not hold too much on other benefits.

Note to @taudac : please make your driver available for that new version.
Can you routinely build your driver at each Hexxeh/rpi-firmware kernel release so that we can build without fearing this hard-linked dependency?

### DIFF
--- a/scripts/raspberryconfig.sh
+++ b/scripts/raspberryconfig.sh
@@ -55,7 +55,7 @@ sudo curl -L --output /usr/bin/rpi-update https://raw.githubusercontent.com/Hexx
 touch /boot/start.elf
 mkdir /lib/modules
 
-KERNEL_VERSION="4.9.41"
+KERNEL_VERSION="4.9.51"
 
 case $KERNEL_VERSION in
     "4.4.9")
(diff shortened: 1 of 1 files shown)


macmpi:
thanks @taudac : please see this https://github.com/volumio/Build/pull/271 to make install more flexible.

Any chance your modules will be integrated in mainstream kernel ?

taudac:
@macmpi Thank you. Integration into the mainstream kernel will take time. For now, I think we can go with [#271](https://github.com/volumio/Build/pull/271).

macmpi:
@volumio without https://github.com/volumio/Build/pull/271 , any Pi kernel update will break TauDAC support.

volumio:
@taudac before merging, you need to make clear that:
- You will responsible of providing sources to anyone interested
- You will relieve Volumio from the obligation to host sources for your contribution

You can just type yes and we will merge

taudac:
yes

volumio:
I reverted it, its now pushed...
As for new kernels, since they introduce every time new problems (or different working mechanisms) we ideally want to stuck with a stable kernel as long as possible and change it ONLY if the new kernel brings very useful features or compatibility with new devices

PR #275, opened 2017-10-28
Improve dynamicswap service definition

macmpi:
Have unit consistent with swap requirements as per [freedesktop.org reference](https://www.freedesktop.org/software/systemd/man/systemd.swap.html#Default%20Dependencies).

Is cleaner for proper reboot & shutdown.
(also similar to what [LibreELEC does](https://github.com/LibreELEC/LibreELEC.tv/blob/master/packages/sysutils/util-linux/system.d/swap.service))

Tested ok on PiZero


volumio:
Thanks, good one

PR #281, opened 2017-11-13
Run CMD in bash and not in user's standard shell

Granjow:
If the user's shell is set to e.g. fish, running the scripts result in an error like:

    Unsupported use of '&&'. In fish, please use 'COMMAND; and COMMAND'.
    fish: which lxpanelctl >/dev/null 2>&1 && lxpanelctl alsastop >/dev/null

This change ensures that the command is run in bash as well, as specified in the shebang line, and not in the custom user shell.

### DIFF
--- a/volumio/opt/vc/bin/dtoverlay-post
+++ b/volumio/opt/vc/bin/dtoverlay-post
@@ -4,7 +4,7 @@ if [ "$DISPLAY" == "" ]; then
 fi
 CMD="lxpanelctl alsastart >/dev/null"
 if [ $EUID -eq 0 ]; then
-    exec su volumio -c "$CMD"
+    exec su volumio --preserve-environment -c "$CMD"
 else
     exec "$CMD"
 fi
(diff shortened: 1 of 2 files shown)


Granjow:
@volumio Any chances to get this merged?

volumio:
What is the purpose of it? Aka which issue solves for yourself?

Granjow:
The one described in the first comment. If a different shell than `bash` is used in `/etc/passwd`, e.g. [fish](http://fishshell.com/), the scripts themselves are executed in Bash due to the shebang line, but the exec command uses the standard shell from `/etc/passwd` again.

So, if I e.g. run `dtparam`, then it only works if I have Bash as standard shell, but it fails with the error in the first comment if I have fish:

    ~> dtparam i2c_arm=on
    Unsupported use of '&&'. In fish, please use 'COMMAND; and COMMAND'.
    fish: which lxpanelctl >/dev/null 2>&1 && lxpanelctl alsastop >/dev/null

Granjow:
@volumio Any news? Or are there issues with the changes?

Granjow:
@volumio ping?

volumio:
Sorry if that took so long!

Granjow:
Thanks for merging!

PR #287, opened 2017-12-11
fix ${LIB_GNUE} expansion

macmpi:
LIB_GNUE already has a leading /
(incidentally noticed, not tested)


volumio:
Should not be a problem

PR #288, opened 2017-12-20
Avoid an error in syslog when there's no wireless hardware

xipmix:
If there is no wireless hardware available, readlink returns an empty string
which causes basename to throw an error.

As far as I can see there's no point running hostapd, so exit.
Exit with an error to prevent dependent systemd units from starting up.

### DIFF
--- a/volumio/bin/hotspot.sh
+++ b/volumio/bin/hotspot.sh
@@ -1,8 +1,19 @@
 #!/bin/bash
 
+INTERFACE=wlan0
+
 case "$1" in
   'start')
-    MODULE=$(basename $(readlink /sys/class/net/wlan0/device/driver/module))
+    modulepath=$(readlink /sys/class/net/${INTERFACE}/device/driver/module)
+    if [ "X" = "X$modulepath" ]; then
+        # Sometimes there is no wireless interface
+        echo "Unable to find driver module name for interface $INTERFACE"
+        echo "Exiting early"
+        exit 1
+    else
+        # Normal case
+        MODULE=$(basename "$modulepath")
+    fi
     ARCH=`/usr/bin/dpkg --print-architecture`
 
     if [ "$MODULE" = "8192cu" ] && [ "$ARCH" = "armhf" ] && !(modinfo "$MODULE" | grep -q '^depends:.*cfg80211.*') ; then


macmpi:
Good catch!
How about more simply just excluding empty `$MODULE` value like below (just change the else condition):
```
  'start')
    MODULE=$(basename $(readlink /sys/class/net/wlan0/device/driver/module))
    ARCH=`/usr/bin/dpkg --print-architecture`

    if [ "$MODULE" = "8192cu" ] && [ "$ARCH" = "armhf" ] && !(modinfo "$MODULE" | grep -q '^depends:.*cfg80211.*') ; then
      echo "Launching Hostapd Edimax"
    /usr/sbin/hostapd-edimax /etc/hostapd/hostapd-edimax.conf
    elif [  "$MODULE" != '' ]; then
      echo "Launching Ordinary Hostapd"
    /usr/sbin/hostapd /etc/hostapd/hostapd.conf
    fi
  ;;

```

xipmix:
My initial motivation for this patch was to avoid the 
```
basename: missing operand
Try 'basename --help' for more information.
```
in the syslog - it was a red herring in trying to understand
the restarts in #1419.

Then it became apparent that something sane needs to be done
if there's no wlan0 interface.

What you suggest is much more compact, but I thought we would want
to explicitly exit with an error so that:
 * hotspot.service returns a failure status
 * which in turn should inhibit dhcp.service and other dependent
   units from starting

N'est ce pas? But I may well be misunderstanding something.

macmpi:
I get your point: indeed my proposed change just removes the error in log message and fixes the erroneous `hostapd` start when no wifi interface is present.

Not sure we really want `hotspot.service` to throw an error in such cases though...
Maybe @volumio can shed some light on how desirable that would be, but indeed, launch logic & links between `hostapd` `dnsmasq` and `dhcpd` might need some tweaking.
Maybe it's the overall service itself that should bear a conditional execution if wlan0 exists.

xipmix:
I did actually test this but I don't recall the outcome in detail,
apart from noting hostap and dhcpd were not running.
I did not think much about how dnsmasq would be affected but
if it is only needed for the hotspot case then I think it's ok.
Probably worth going through the boot log for with and without
the patch and adding a summary to the comments here.

volumio:
I agree to the fact that we should launch hostapd & friends only if there is wlan0, so instead of editing this file we should prevent hotspot.sh from starting at all if there is no wlan0

xipmix:
So after looking at this for a while I see that hotspot and friends are started in two places:
 * wireless.service calls wireless.js, which unconditionally tries to start hotspot.service
 * systemd also tries to bring up hotspot.service

wireless.service wants to be ```Before: volumio.service```, but afaict there is no dependency expressed in the unit files between wireless.service and hotspot.service.

My thought for what to do here
 * add code to wireless.js that allows it to exit without trying to start anything if there's no wlan interface
 * add a directive to hotspot.service that makes it not start until wireless.service has started (```After: ?```), so if wireless.service fails, hotspot.service does not start.

In the case of wlan0 existing, I thought it would be nice if one could call wireless.js in two ways: 'start' which just starts the wireless as a wifi client and 'hotstart' which starts in hotspot mode. But I'm not sure that's necessary.

xipmix:
superceded by #297 + #303, closing.

PR #289, opened 2017-12-20
Add tv-service alias

Saiyato:
I hope this helps for the Kodi plugin


macmpi:
Can't this be done at runtime in plugin `onVolumioStart` routine, or at install in `install.sh`?

volumio:
Yes, should be. But I find it useful to have tvservice in alias anyway.
What was your concern on this?

macmpi:
Thought `tvservice` was of very little use in Volumio2 baseline config with `gpu_mem=16` (and therefore implied cut-down `start_cd.elf`), so was surprised to see it coming into baseline stuff.

Probably makes sense for video plugins, but one would expect those would make their own relevant adjustments, including allowing more `gpu_mem`, etc...

PR #296, opened 2018-02-05
Buid scripts for FriendlyArm NanoPi-Neo, -Neo Air, Neo2 devices

nikkov:
Platform files will be download from my repository


volumio:
Thanks Nik!

PR #297, opened 2018-02-08
Prevent systemd starting hotspot at boot

xipmix:
This addresses an issue noticed in #288, that we want hotspot.service to be controlled by wireless.js alone. Allowing systemd to start hotspot.service independently just makes things harder to debug,

### DIFF
--- a/scripts/volumioconfig.sh
+++ b/scripts/volumioconfig.sh
@@ -460,6 +460,10 @@ echo "Setting Mpd to SystemD instead of Init"
 update-rc.d mpd remove
 systemctl enable mpd.service
 
+echo "Preventing hotspot services from starting at boot"
+systemctl disable hotspot.service
+systemctl disable dnsmasq.service
+
 echo "Preventing un-needed dhcp servers to start automatically"
 systemctl disable isc-dhcp-server.service
 systemctl disable dhcpd.service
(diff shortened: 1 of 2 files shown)


volumio/bin/firststart.sh
```diff
@@ -15,6 +15,11 @@ rm -v /etc/ssh/ssh_host_*
 echo "Generating SSH host keys"
 dpkg-reconfigure openssh-server
 
+echo "Ensuring hotspot startup is controlled only by wireless.js"
```
volumio: This should be removed, and we should not enable those services in the build script

xipmix: Thanks for that. I found a better place to do this, not sure if the firstboot.sh change is needed at all now.

volumio:
Can you also remove the new lines on firststart.sh?

xipmix:
Sure, done now. I have tested the disabling works on an existing system but I have not done a full build and tested that.

xipmix:
Cleaned up and rebased.

xipmix:
I built an image to test with this patch applied. At firstboot, hotspot runs (as expected). Then I disabled it in the UI and rebooted; system came up without hotspot. Then I enabled it in the UI and it started, finally disabled again and it stopped.

Issue #291, opened 2018-01-18
TP-LINK AC 600 Archer T2UH seems to be not supported


thomasdoerr:
Hi, I commented in this post https://github.com/volumio/Volumio2/issues/1452 that my volumio installation (after the update from 2.246 to the latest 2.348) freezes, if my TP-LINK AC 600 Archer T2UH was plugged in. In the moment when I plug it in, volumio freezes when I try to fire a command. I noticed that, since nothing worked after the update.

xipmix asked me to open an issue here. 

My pi is now running without the TP-LINK. One of the main raisons why I used the TP-LINK was, that channel 12/13 were not enabled on the pi and my router sometimes used these channels. 
Today, probably after installing "http://www.fars-robotics.net/install-wifi -O /usr/bin/install-wifi", both channels are active. So everything is fine for me now, but there might be others having the same problem.
Regards, Thomas

macmpi:
Strange...Volumio build is indeed supposed to install MrEngman drivers...

Can you give output of `lsmod `and `lsusb` when you have the dongle plugged and working?
Can you show `dmesg` output when you plug the dongle on working / non-working setup?

thomasdoerr:
hmm, as I wrote, when I plug in the dongle, I'm unable to fire any command. I fixed my machine, so I need to recreate the situation. I'll try to reproduce it. Give me a little bit of time :-)

macmpi:
if you can already provide info on working setup, that good.
Apparently that beast uses mt7610 and Volumio script pulls the right driver file (that exists)

thomasdoerr:
ok

macmpi:
In the meantime, I checked v2.348 image file, and can confirm all drivers and firmware files are where they should (in particular for mt7610).

Could you check if your USB dongle works with a newly imaged v2.348 SD please?
Maybe something went wrong in the upgrade due to previous file modifications in original install...

thomasdoerr:
I checked it with a fresh image, this is always my first check, if I have update problems. Same issue.
But now I managed to access volumio via ssh in both cases with plug dongle.
You'll find two text files with (hopefully) the information you asked. In the screenshot you see, that in the case of 2.348 plus TP-LINK, when I try to call sudo journalctl -f it hangs.

Hope you'll be able to find the bug! :-)


[tp-link-issue.zip](https://github.com/volumio/Build/files/1642688/tp-link-issue.zip)

macmpi:
[...]

PR #300, opened 2018-02-11
Pi: update MrEngman drivers location

macmpi:
as per [this](https://www.raspberrypi.org/forums/viewtopic.php?f=28&t=62371&p=1271993#p1271965).
Should in particular fix https://github.com/volumio/Build/issues/291 for good.


volumio:
Thanks!

volumio:
What about:
1- Adding all possible drivers, including mt7612
2- Doing a loop as you suggest, so it's nicer to see and easier to mantain
3- Copy once in a while the zips into a github repo (where we provide infos, like a json file that we parse in the script so everytime we add an archive we don't need to update the script). This is also to avoid borrowing bandwith to mr engman server
I've created a repo, and added you as collaborator, if you like the idea:
https://github.com/volumio/wifi-firmwares-pi

PR #304, opened 2018-02-14
Pi: improve MrEngman wifi drivers install

macmpi:
Now with all MrEngman drivers (i.e. add mt7612), and reducing code duplication with a loop on drivers.

Still uses MrEngman repo, awaiting for eventual local mirror structure.


volumio:
Very good.
I think we should indeed put those files in our repo:
https://github.com/volumio/wifi-firmwares-pi

By keeping the same file structure, we just change to raw.github etc

Also I think it would be critical to get sources, and maybe start building our own...

macmpi:
ok will pull some files in repo then.
Yes building them would be nice, but testing might require lots of HW.
Also we should then include a reference `buildroot` environment in build script so that it can be built by anyone (note this could apply to other custom-built components like `mpd` & al.).

BTW, PINN (enhanced NOOBS) does build quite a [number of USB dongle drivers](https://github.com/procount/pinn/blob/master/changes.md#additional-network-drivers)...Maybe build environment and sources could be derived from there...

volumio:
Good hint about PINN, if he provides both sources and compiled packages then it's a no brainer, let's go with this.

What do you think?

PR #305, opened 2018-02-14
Use sudoers.d directory for custom permissions

piffio:
It is a good practice to leave /etc/sudoers to the official distro
packages, and then create specific files under /etc/sudoers.d for
adding custom configurations.
This ensures that in case of major update of the sudoers file provided
by the official package we don't lose our custom rules while still
benefitting from any major changes happening upstream.

For Volumio, this means creating a /etc/sudoers.d/volumio-user file with
all the relevant configuration.


xipmix:
Looks good to me, can you confirm you tested this? Might also help with volumio/Volumio2#1247

piffio:
Yes, I did test it on a build for Raspberry, and everything worked out of the box.
Thanks for merging.

PR #313, opened 2018-03-16
include all latest brcm firmware

macmpi:
latest support for Pi3/Pi0W and Pi3+


macmpi:
This is safe on 4.14 context will all new core firmware release as in this branch.

May have some instabilities on 4.9, unless we "borrow" supposedly fixed 4.9 core firmware files & drivers from 2018-03-13 [Raspbian archive](https://www.raspberrypi.org/downloads/raspbian/) (unfortunately 4.9 is not built anymore on https://github.com/Hexxeh/rpi-firmware/)

PR #314, opened 2018-03-16
udev filter for Pi3+ lan78xx

macmpi:
gives precedence to connected USB-ethernet interface over built-in ethernet.
Now support all Pis with built-in smsc95xx or lan78xx
(not tested on Pi3b+)

### DIFF
--- a/scripts/raspberryconfig.sh
+++ b/scripts/raspberryconfig.sh
@@ -184,8 +184,8 @@ ln -s /opt/vc/lib/libvchiq_arm.so /usr/lib/libvchiq_arm.so
 ln -s /opt/vc/bin/vcgencmd /usr/bin/vcgencmd
 ln -s /opt/vc/lib/libvcos.so /usr/lib/libvcos.so
 
-# changing external ethX priority rule for Pi as built-in eth _is_ on USB (smsc95xx driver)
-sed -i 's/KERNEL==\"eth/DRIVERS!=\"smsc95xx\", &/' /etc/udev/rules.d/99-Volumio-net.rules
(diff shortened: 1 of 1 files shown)


volumio:
Got it, one question. Why would someone use an external ethernet?

macmpi:
In case internal fried for instance :) (or had other benefits: better performance, etc)
Just made it originally to pair with similar feature with wifi side.

if you want to test without rebuilding whole, target file is:
```
# Give Volumio interface priority to external wlan or eth USB network dongles if connected
ACTION=="add", SUBSYSTEM=="net", SUBSYSTEMS=="usb", ENV{DEVTYPE}=="wlan", ATTR{type}=="1", KERNEL=="wlan[1-9]*", RUN+="/bin/rename_netiface0.sh '%E{INTERFACE}'"
ACTION=="add", SUBSYSTEM=="net", SUBSYSTEMS=="usb", ENV{DEVTYPE}!="wlan", ATTR{type}=="1", DRIVERS!="smsc95xx", DRIVERS!="lan78xx", KERNEL=="eth[1-9]*", RUN+="/bin/rename_netiface0.sh '%E{INTERFACE}'"
```

macmpi:
> Got it, one question. Why would someone use an external ethernet?

how about [this](https://volumio.org/raspberry-pi-3-b-plus-audio-review/)? ;)
> So, for now, if you avoid using Ethernet to retrieve your Hi-Res files, the PI 3B+is just perfect to feed your USB DAC with bit-perfect joy.

Any chance you could test with some asix-based dongle (like ax88179 USB3 ones)?

PR #316, opened 2018-03-22
get firmware & kernel from latest stable

macmpi:
Hopefully working: supposedly same as 2018-03-13 Raspbian
Note: MrEngman drivers may not be available yet, Taudac may need to be updated too..


volumio:
You did it! It works! Great !

macmpi:
Sweet!
Seems MrEngman is working on compiling his drivers.
Might be safer if @taudac could also recompile with that slightly updated kernel version...

taudac:
Can't find the right kernel source, firmware commit 936a8dc3 points to kernel commit 921b6254. The later doesn't exist.

macmpi:
Yes: it's a bit of a hack-ish version: see https://github.com/raspberrypi/firmware/issues/954
On the more standard release front, please note 4.14.29 is out.

taudac:
I have recompiled the modules. They are binary identical to my original "for-4.9.80" release.

macmpi:
Thanks for checking and confirmation.

While I understand upstreaming your driver might be a long process, would you be considering submitting directly to https://github.com/raspberrypi/linux as a first step?
It very much seems other DAC vendors did this, and got integrated pretty fast.

PR #317, opened 2018-03-22
Pi3b+: try 4.14.29

macmpi:
might fix eth0 issues...


volumio:
Nope unfortunately...

macmpi:
Interesting...
It seems it does within berryboot kernel, for several distributions.

gkkpch:
Missing firmware perhaps? Kernel versions > 4.12 don't have any. You must add your own.

PR #331, opened 2018-04-12
depth 1 for cloning platform files

cdu13a:
Fetching the whole git history for platform files seems unnecessary. Unless there is a reason for current behavior, It's probably better to use --depth 1 when we fetch the platform files.


volumio:
Thanks

PR #337, opened 2018-05-20
fix error alsa bufer for Amlogic audio

150balbes:


150balbes:
can see the details of the fix here.

https://volumio.org/forum/volumio-box-amlogic-s905-s905x-s912-s802-s812-s805-t8028-50.html

PR #356, opened 2018-09-19
4.14.70 commit fix for kernel panics

macmpi:
as per https://github.com/volumio/Build/commit/c00a19098320053824010d132d6bfb39774a7d2f#commitcomment-30554050


volumio:
Is it safe to merge? Even if it's not tied to a version bump? Will we have troubles with headers and uname string?

macmpi:
I do not expect any issue, but can not test right now.
rpi-update just needs a commit number: we did similar things in the past.
Kernel rev is unchanged (still 1144)

macmpi:
If you do not feel confortable, just jump to 4.14.71 in a few bit: it potentially contains a major USB fix (awaiting final confirmation) ;)
https://github.com/Hexxeh/rpi-firmware/commit/c919d632ddc2a88bcb87b7d0cddd61446d1a36bf

volumio:
Done https://github.com/volumio/Build/commit/19491f50eded236de6874d31640dedebc248dfa0

PR #361, opened 2018-10-27
Update to stable Node 8 LTS

piffio:
Node 8.12.0 includes some minor improvements but most importantly
quite a few security fixes compared to 8.11.1.

Not very easy to test since it requires the specific node version and pre-built node modules to be already available on the update server


volumio:
I agree with you, however, given that node 10 is going into LTS starting from the end of this month, I'll be more inclined to switch to node 10 (starting mid of next month).
In case we'll find breaking changes with node 10, we can go with 8.2

Issue #347, opened 2018-07-30
System Update does not preserve /boot/config.txt changes (crossref)


xipmix:
Cross ref to volumio/Volumio2#1568

gkkpch:
The problem is the following code in scripts/initramfs/init:
`if [ -e "/mnt/imgpart/config.txt.bak" ]; then

    print_msg "Restoring custom config.txt content"
    I2S=`sed -n -e '/#### Volumio i2s setting below: do not alter ####/,$p' /mnt/imgpart/config.txt.bak`
    echo "" >> /boot/config.txt
    echo "$I2S" >> /boot/config.txt
    rm /mnt/imgpart/config.txt.bak

  fi`

As shown, it does not really restore config.txt from a backup, it just takes the DAC DT overlay settings from the old config and adds it to the one from the update.
There is no perfect way to solve this (at least, I don't see one yet).
Option 1, taking the update, will always preserve a bootable system as it was tested that way), but some customisation settings could be missing (new or changed params).
Option 2, using the backup settings, will only work, as long as there was no need in the new update to add or change a param setting. Therefore this has potential for boot failure.

And improvement of option 1 could be a merge: all parameter settings from config.txt.bak are added to the end of config.txt in case they were not already present there.
This would automatically add the DAC settings, no need to search with "sed" and add them.
However, this option is not 100% proof in cases where a value was changed during customising or because the new volumio version required it to be.

swinokur:
You could do option #1, but you should then move config.txt to another name (config.txt.saved, etc). Currently the existing config.txt is backed up during the first phase of upgrade, but then for some reason the backup is deleted. So, if someone has made configurations and not remembered / recorded what they did, then those changes will be lost through the upgrade procedure.

gvolt:
What do you think about that:

As recently posted in the [Volumio forum](https://volumio.org/forum/config-txt-overwritten-each-system-update-t10667.html#p53111) there is already an "include" option for the config.txt ([https://www.raspberrypi.org/documentation/configuration/config-txt/misc.md](https://www.raspberrypi.org/documentation/configuration/config-txt/misc.md)). The content of the file specified will then be inserted into the config.txt. This of course would require Volumio incorporating the "include" option into the default config.txt file.

In order to ensure a safe reboot after an update the include option could be commented out at the first boot. On the dev page of the webinterface a switch could be added to activate (uncomment) the include option in the config.txt file.

gkkpch:
[...]

PR #368, opened 2019-01-24
Solution to issue #347 and #1568

gvolt:
I suggest to add the include option `#include userconfig.txt` to /boot/config in order to solve issue #347 "System Update does not preserve /boot/config.txt changes (crossref)" respectively [#1568](https://github.com/volumio/Volumio2/issues/1568).

The user would have to fill the predetermined file (/boot/userconfig.txt) with his additions and remove the comment statement (#) from the include option in /boot/config.txt.

On an update the include option would be commented out again (by re-applying the vanilla config.txt) but the additions in userconfig.txt would be preserved. That way the update process would not have to deal with backing up and restoring user made changes and can remain unchanged. Furthermore it would be assured that the system boot will not fail because of user made changes to config.txt. Instead the user has the opportunity to check if the content of the userconfig.txt is still appropriate. If that's the case the include option in /boot/config just has to be uncommented again.

### DIFF
--- a/scripts/raspberryconfig.sh
+++ b/scripts/raspberryconfig.sh
@@ -168,7 +168,9 @@ dtparam=audio=on
 audio_pwm_mode=2
 dtparam=i2c_arm=on
 disable_splash=1
-hdmi_force_hotplug=1" >> /boot/config.txt
+hdmi_force_hotplug=1
+
+#include userconfig.txt" >> /boot/config.txt
 
 echo "Writing cmdline.txt file"
 echo "splash quiet plymouth.ignore-serial-consoles dwc_otg.fiq_enable=1 dwc_otg.fiq_fsm_enable=1 dwc_otg.fiq_fsm_mask=0xF dwc_otg.nak_holdoff=1 console=serial0,115200 kgdboc=serial0,115200 console=tty1 imgpart=/dev/mmcblk0p2 imgfile=/volumio_current.sqsh elevator=noop rootwait bootdelay=5 logo.nologo vt.global_cursor_default=0 loglevel=0" >> /boot/cmdline.txt
--- a/scripts/raspberryconfig.sh
+++ b/scripts/raspberryconfig.sh
@@ -170,7 +170,7 @@ dtparam=i2c_arm=on
 disable_splash=1
 hdmi_force_hotplug=1
 
-#include userconfig.txt" >> /boot/config.txt
+include userconfig.txt" >> /boot/config.txt
 
 echo "Writing cmdline.txt file"
 echo "splash quiet plymouth.ignore-serial-consoles dwc_otg.fiq_enable=1 dwc_otg.fiq_fsm_enable=1 dwc_otg.fiq_fsm_mask=0xF dwc_otg.nak_holdoff=1 console=serial0,115200 kgdboc=serial0,115200 console=tty1 imgpart=/dev/mmcblk0p2 imgfile=/volumio_current.sqsh elevator=noop rootwait bootdelay=5 logo.nologo vt.global_cursor_default=0 loglevel=0" >> /boot/cmdline.txt
--- a/volumio/bin/firststart.sh
+++ b/volumio/bin/firststart.sh
@@ -8,6 +8,9 @@ dpkg --configure --pending
 echo "Creating /var/log/samba folder"
 mkdir /var/log/samba
 
+echo "Creating /boot/userconfig.txt"
+echo "# Add your custom config.txt options to this file, which will be preserved during updates" >> /boot/userconfig.txt
+
 echo "Removing default SSH host keys"
 # These should be created on first boot to ensure they are unique on each system
 rm -v /etc/ssh/ssh_host_*


volumio:
@gvolt I like the idea, and I would suggest to:
- Do not have the comment by default: in case an update fails, the user will check it
- Have the userconfig.txt created on first boot (adding to firststart.sh). Ideally userconfig.txt should have a line explaning what it does. Like #Add in this file your custom config.txt options, which will be preserved during updates.
We have to do it in first config, otherwise the file will be overwritten on updates.

What do you think?

gvolt:
@volumio I fully agree with your suggestions. :)

volumio:
Brilliant! Thanks!

misko903:
this is good idea. but where is the file userconfig.txt located?

misko903:
Thanks for this. Hopefully I dont need to Delete user data all the time before update.
I am using custom IR and mpd_oled display.

Saiyato:
Hi @gvolt ,

I've been making my plugins a bit more robust, especially Kodi relies on gpu memory settings. However if I move the gpu_mem* settings to userconfig.txt and include it, they seem not to be loaded. Do you have any experience in this regard?

What I did:
- create userconfig.txt and include it into config.txt
- move the gpu settings to userconfig.txt (gpu_mem_1024=256)
- read the gpu mem using: vcgencmd get_mem gpu

Otherwise I'll just leave the gpu memory settings in the /boot/config.txt file instead.

gvolt:
Hi @Saiyato ,

does /boot/config.txt contain the line `include userconfig.txt`? I am asking because if I remember correctly using the full path to userconfig.txt (i.e. `include /boot/userconfig.txt`) would not work. Therefore userconfig.txt has to be placed in /boot, too.

You may also check if there is a gpu_mem_1024 setting _after_ the `include userconfig.txt` statement in /boot/config.txt. That would override the setting in the userconfig.txt.

Saiyato:
Hi @gvolt ,

Thanks for the quick response, below my /boot/config*.txt:

```
volumio@volumioserre:~$ cat /boot/config.txt
include userconfig.txt

initramfs volumio.initrd
gpu_mem=32
max_usb_current=1
dtparam=audio=on
audio_pwm_mode=2
dtparam=i2c_arm=on
disable_splash=1
hdmi_force_hotplug=1

# GPU memory settings

#gpu_mem_1024=256
#gpu_mem_512=144
#gpu_mem_256=112

volumio@volumioserre:~$ cat /boot/userconfig.txt
# GPU memory settings

gpu_mem_1024=256
gpu_mem_512=144
gpu_mem_256=112
hdmi_force_hotplug=1

volumio@volumioserre:~$ ls -al /boot/*config.txt
-rw-rw-rw- 1 root root 245 Jan  4 16:05 /boot/config.txt
-rw-rw-rw- 1 root root  95 Jan  4 13:44 /boot/userconfig.txt
```
Querying the memory gives me:
```
volumio@volumioserre:~$ vcgencmd get_mem gpu
gpu=32M
```
It looks in order, I just can't wrap my head around it as to why it's not working.

gvolt:
Although gpu_mem should not overide gpu_mem_1024 I would suggest to move `include userconfig.txt` to the end of /boot/config.txt and see if that helps.

Saiyato:
Unfortunately that yields the same results, somehow I thing the inclusion is not working for the gpu mem part, maybe the concatenation only happens after the memory settings have been loaded.

According to this site: https://www.raspberrypi.org/documentation/configuration/config-txt/README.md

> This is read by the GPU before the ARM CPU and Linux are initialised.

The GPU reads first, maybe it's impossible to concatenate at that point. Updating the HDMI_DRIVE setting did work, so it's working for some settings, but not for others.

Should I just leave the ones that aren't working with userconfig in the config.txt file instead?

gvolt:
Hm, could be - would be a bummer... Did you find other settings which also don't work if they are placed in the userconfig.txt?

Saiyato:
Not yet, but to be honest I was just working on the Kodi plugin, it was in dire need of some cleanup ;)

gvolt:
It seems gpu_mem* settings in userconfig.txt work (only) if there is no such setting in /boot/config.txt.

gvolt:
To add to my previous post: I commented the gpu_mem setting in /boot/config.txt and rebooted with a gpu_mem setting in /boot/userconfig.txt. That way the setting gets followed. The same goes for e.g. gpu_mem_1024 :)

Saiyato:
I just commented the gpu_mem in /boot/config.txt, but I'm only getting 76M when querying the memory.

```
volumio@volumioserre:~$ cat /boot/config.txt
initramfs volumio.initrd
#gpu_mem=32
max_usb_current=1
dtparam=audio=on
audio_pwm_mode=2
dtparam=i2c_arm=on
disable_splash=1
hdmi_force_hotplug=1

include userconfig.txt

volumio@volumioserre:~$ cat /boot/userconfig.txt
# GPU memory settings
gpu_mem_1024=256
gpu_mem_512=144
gpu_mem_256=112

hdmi_force_hotplug=1
hdmi_drive=1

volumio@volumioserre:~$ vcgencmd get_mem gpu
gpu=76M
```

When adding gpu_mem_1024 to /boot/config.txt I do get the correct values. I will try to find out why, because I want to move the config to userconfig.txt. I'm using a Pi2B btw.

gvolt:
@Saiyato @volumio 
Today I have been asking and discussing the issue of placing the gpu_mem option in an included file like userconfig.txt over [here](https://github.com/raspberrypi/firmware/issues/1332).

Very helpful and informative thread, but sad news are: gpu_mem does indeed not work in files referenced by the `include` option in /boot/config.txt.

What are your thoughts about making userconfig.txt the file that gets reset on each update and let config.txt be untouched?

Issue #348, opened 2018-08-18
Build failed on ubuntu 18.04


howardqiao:
Get:1 http://archive.volumio.org/raspbian jessie InRelease [14.9 kB]
Err:1 http://archive.volumio.org/raspbian jessie InRelease
  The following signatures couldn't be verified because the public key is not available: NO_PUBKEY 9165938D90FDDD2E
Reading package lists... Done
W: GPG error: http://archive.volumio.org/raspbian jessie InRelease: The following signatures couldn't be verified because the public key is not available: NO_PUBKEY 9165938D90FDDD2E
E: The repository 'http://archive.volumio.org/raspbian jessie InRelease' is not signed.
N: Updating from such a repository can't be done securely, and is therefore disabled by default.
N: See apt-secure(8) manpage for repository creation and user configuration details.
......
W: Target Translations (main/i18n/Translation-en) is configured multiple times in /home/q/aoide-dac-drivers/build/build/arm/root/etc/apt/sources.list.d/multistrap-assets.list:1 and /home/q/aoide-dac-drivers/build/build/arm/root/etc/apt/sources.list.d/multistrap-utils.list:1
apt update failed. Exit value: 100

gkkpch:
try removing “build/arm/root/etc/initramfs/hooks” and rebuild the image with - d and -v only.
No need for the -b parameter, otherwise it rebuilds the rootfs too and you’re back at the beginning.

gkkpch:
Sorry, first read the README.md, you might have to edit the multistrap script, instructions are there.
Then start from the beginning and when you get an error while building the image, do the above.

howardqiao:
Sorry,but there's no "$forceyes" in /usr/sbin/multistrap
But i can build the image on ubuntu 16.04 with below parameters.
sudo ./build.sh -b arm -d pi -v 2.344

gkkpch:
Then please guys, post a complete log so I can help and do ‘t have to guess. A partial log is,  well, dangerous as you might have missed something you thought was unimportant. And “Same here” does not help at all, sorry.
I have a Ubuntu Bionic (18.04?) basic server  and seem to have a working setup...

gkkpch:
OK, I have been able to reproduce this on a fresh Ubuntu install.
No idea what is causing this.
I did noticed the version on Ubuntu uses a Parameter APT_CONFIG not seen on Debian jessie, but I'm not sure if there is a relation. Unfortunately I have practically no time to follow it up at the moment.
[...]

PR #369, opened 2019-04-02
Exit build script on failures

ashthespy:
Especially `multistrap`!
After quite some head scratching as to why my images weren't booting, scrutinised the build process in more detail to discover multistrap was failing. 
This would make errors like #348 easier to catch.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -325,6 +325,11 @@ case "$DEVICE" in
     check_os_release "armv7" "$VERSION" "$DEVICE"
     sh scripts/nanopineoimage.sh -v "$VERSION" -p "$PATCH" -a armv7
     ;;
+  "") echo 'No device specified'
+    ;;
+  *) echo Unknown/Unsupported device: $DEVICE
+    exit 1
+    ;;
 esac
 
 #When the tar is created we can build the docker layer
--- a/build.sh
+++ b/build.sh
@@ -6,6 +6,7 @@
 #
 # Dependencies:
 # parted squashfs-tools dosfstools multistrap qemu binfmt-support qemu-user-static kpartx
+set -eo pipefail
 
 #Set fonts for Help.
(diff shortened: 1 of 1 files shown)


gkkpch:
Agree, but checking the log is still strongly adviced.
I have seen other errors, like Debian repo's not responding, leading to similar errors.
As for the multistrap issue, this is known from Ubuntu and Stretch OS hosts.
Meanwhile the multistrap issue has been resolved, but an issue with building our initramfs is still being debugged, unfortunately on hold because of other priorities. 

Only Debian jessie is guaranteed to work as a host OS at the moment.

ashthespy:
>Agree, but checking the log is still strongly adviced.
>I have seen other errors, like Debian repo's not responding, leading to similar errors.
>Only Debian jessie is guaranteed to work as a host OS at the moment.

That explains a lot - I tried first with stretch, and then buster. Will try again with Jessie.

gkkpch:
Thought so, another stretch/ buster or Ubuntu user....
And we already have a number of users preferring Ubuntu for their build environment.
Seems this slowly needs more attention.... moved #348 to 3rd on my todo list.

ashthespy:
@gkkpch Don't mean to clutter this PR, I am having quite the hard time even with `jessie` now.
Something about the wrong binary format of `/bin/bash` 

Is there documentation somewhere for the ideal build environment?  I don't recall it being so hard to build this about a year ago! :=D

gkkpch:
just email directly

gkkpch:
use volumio(at)bluewin.ch

ashthespy:
@volumio fe83e03 has introduced conflicts via a reference to a `motivo` device. Seems like a mistake as 
 `scripts/motivoimage.sh` wasn't added to the repo?

PR #372, opened 2019-04-11
Fixes in `volumioconfig.sh`

ashthespy:
Some warnings and fixes found by using `set -e`. 
Maybe a good idea to turn it on even in this script?


volumio:
Thanks! Why would you turn it off?

ashthespy:
>Thanks! Why would you turn it off?

It's not on currently - see #369

PR #383, opened 2019-07-08
Blacklisting unwanted USB wifi drivers

macmpi:
rtl8192cu and rtl8xxxu come with 4.19.x kernel

### DIFF
--- a/scripts/raspberryconfig.sh
+++ b/scripts/raspberryconfig.sh
@@ -126,6 +126,10 @@ cp /boot/.firmware_revision /boot/.firmware_revision_kernel
 echo "Updating bootloader files *.elf *.dat *.bin"
 echo y | SKIP_KERNEL=1 WANT_PI4=1 SKIP_CHECK_PARTITION=1 UPDATE_SELF=0 /usr/bin/rpi-update $FIRMWARE_COMMIT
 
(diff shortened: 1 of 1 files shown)


volumio:
Thanks, can we add a version check before those? 
I mean: if kernel version is >= 4.19 then blacklist them 

What do you think?

macmpi:
Hi.
Your call: don't know how much you intend to support/roll-back to older kernels: anyway these would not cause issue with earlier kernels either...
(BTW about older kernels: could we not remove the older ones and linked stuff?)

volumio:
We need older kernel switches for some OEM builds ;) 
To this regard that's why I asked for the switch as well ;)

macmpi:
Comparing version numbers is a bit tricky... (and those blacklists cause no harm to previous Kernels)

Would [this switch](https://stackoverflow.com/questions/16989598/bash-comparing-version-numbers) make sense to you? (second one, without sort -V)?

macmpi:
:)  forgot I did submit that a while back.
We inherit those blacklists from `raspberrypi-sys-mods` anyway, so closing.

Note though, `rtl8192cu` seems better for AP mode under 4.19...and may come back depending on https://github.com/RPi-Distro/raspberrypi-sys-mods/issues/37

PR #390, opened 2019-10-02
SKIP_CHECK_PARTITION now merged

macmpi:
BTW how about installing `rpi-update` the `apt` way?

### DIFF
--- a/scripts/raspberryconfig.sh
+++ b/scripts/raspberryconfig.sh
@@ -59,9 +59,7 @@ apt-get -y install binutils i2c-tools
 #apt-get -y install libnewt0.52 whiptail triggerhappy lua5.1 locales
 
 echo "Installing Kernel from Rpi-Update"
-# Temporary fix for PI4 check partition, until it gets merged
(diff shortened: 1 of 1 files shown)


ashthespy:
General question (also to @macmpi as you seem to know the tricks of the Rpi's ;-) )
Is there a reason not to use the [`ROOT_PART`/`BOOT_PART`](https://github.com/Hexxeh/rpi-update#root_path-and-boot_path) to copy files into the rootfs, instead of running the script from chroot?

ashthespy:
Is this still valid or can it be closed?

macmpi:
Have not tested lately
It seems @volumio ´s cached version still differs from original hexxeh repo.
Would probably be worth syncing (more future proof), since hexxeh adapted WANT_PI4 SKIP_CHECK_PARTITION a while ago.

Also using apt package rather then curl to pull it would probably be cleaner too

ashthespy:
Closing this as I don't think we are going to be doing anything radical with the master (Jessie) branch going forward. 
[volumioOS branch](https://github.com/volumio/Build/blob/0804b7881fd390d7687d5e6424ed71036810f130/recipes/devices/pi.sh#L179) uses the upstream `rpi-update` already.

PR #393, opened 2019-10-08
Revert "Prepare for Debian Buster"

gkkpch:
Reverts volumio/Build#392


basnijholt:
Why was this reverted?

gkkpch:
because we decided to keep it in a separate branch (buster-new).
It is x86-only at the moment, which boots fine but still lacks a number of packages. It is WIP!

PR #400, opened 2020-01-30
Configure a Samba password for volumio user

BenjaminDebeerst:
Samba uses system users for access control, but uses independent
passwords. This configures a default password for the volumio user upon
system configuration.

This PR is split out from functionality I originally put in volumio/Volumio2#1883, for rationale and discussion, see there. The two PRs are not dependent on each other though, and can thus be merged independently.

**Note:** This adds the volumio default password ("volumio") as cleartext to the codebase. That's not ideal, but given the security of the default password not such a big problem. If that is undesirable, I could investigate whether it is possible to provide the password hash or even the complete samba password DB as a file (instead of using `smbpasswd`).


BenjaminDebeerst:
Closing for staleness.

PR #408, opened 2020-03-29
Update volumioconfig.sh

balbuze:
add volumio user in the group "systemd-journal" in order to read journalctl from a plugin.

### DIFF
--- a/scripts/volumioconfig.sh
+++ b/scripts/volumioconfig.sh
@@ -62,7 +62,7 @@ echo ""
 #Adding Main user Volumio
 echo "Adding Volumio User"
 groupadd volumio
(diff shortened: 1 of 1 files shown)


volumio:
Well done!

balbuze:
this seems to be not used even in v 2.756.
what is the problem? or where to put this mod to have it a next build?

volumio:
If I type groups volumio, I get:
volumio : volumio adm lp dialout cdrom floppy audio dip video plugdev netdev i2c input systemd-journal gpio spi

Probably something you've installed (a plugin?) did overwrite the groups file and this failed to update. Can you check?

PR #409, opened 2020-04-17
Pi: update brcm firmware & improve BT coexistence

macmpi:
https://github.com/RPi-Distro/firmware-nonfree/issues/8#issuecomment-615339218
(need to get file from https://archive.raspberrypi.org/debian/pool/main/f/firmware-nonfree/firmware-brcm80211_20190114-1+rpt6_all.deb )


volumio:
Were you able to test it with your PI?

macmpi:
No, unfortunately as I do not have a Pi4, just PiZeroW.
Reports in original thread seem rather positive.
Firmware files were quite behind anyway (+rpt2), in particular without kr00k patches

volumio:
in your PIZerow works?

macmpi:
Note latest NVRAM param tweak is on brcmfmac43455, not brcmfmac43430 of PiZeroW

My PiZeroW doing Wifi & BT is on a Alpine build (mpd audio out via bluez-alsa), not Volumio, but works good with those firmware files.

PR #410, opened 2020-04-25
moved 'rm -rf .git' after 'git rev-parse HEAD'

makuhama:
`VOLUMIO_FE_VERSION` in `/etc/os-release` was empty.

Reason:

Folder `/volumio/http/www/.git` was deleted before `git ... rev-parse HEAD`

So I moved the deletion after writing version information to `/etc/os-release`.

In the wake of that I also added `rm -rf build/$BUILD/root/volumio/.git` saving another 29MiB from the image

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -186,8 +186,6 @@ if [ -n "$BUILD" ]; then
   git clone --depth 1 -b dist --single-branch https://github.com/volumio/Volumio2-UI.git "build/$BUILD/root/volumio/http/www"
   echo 'Cloning Volumio3 UI'
   git clone --depth 1 -b dist3 --single-branch https://github.com/volumio/Volumio2-UI.git "build/$BUILD/root/volumio/http/www3"
-  rm -rf build/$BUILD/root/volumio/http/www/.git
-  rm -rf build/$BUILD/root/volumio/http/www3/.git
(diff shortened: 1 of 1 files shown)


ashthespy:
Umm, I thought I fixed this in 2c26631fe6afa9e01d8447208eb23d25e4e5b8ea 


>In the wake of that I also added rm -rf build/$BUILD/root/volumio/.git saving another 29MiB from the image

I believe that is there on purpose - it helps with the [`volumio pull` commands](https://github.com/volumio/Volumio2/blob/96afe745de52b0c37801df309ae9bfb72fe1f635/app/plugins/system_controller/volumio_command_line_client/volumio.sh#L44-L48)

ashthespy:
Ah, looks like ef5da971c6de39c271b162596aa09378bec8523f broke it..

volumio:
Thanks for fixing it. As @ashthespy points out we need to retain the /volumio/.git folder , so please remove the line when you remove this .git folder.
29MB are a really great gain in size, let's explore other options to achieve the same result even if we should have taken all possible steps to reduce the size of it via https://github.com/volumio/Build/blob/master/build.sh#L179

volumio:
Thanks! Why did you close it? it was fine ;)

PR #411, opened 2020-04-25
moved 'rm -rf .git' after 'git rev-parse HEAD'

makuhama:
`VOLUMIO_FE_VERSION` in `/etc/os-release` was empty.

Reason:

Folder `/volumio/http/www/.git` was deleted before `git ... rev-parse HEAD`

So I moved the deletion after writing version information to `/etc/os-release`.


volumio:
Thanks! Very welcome! Feel free to suggest other impros

PR #416, opened 2020-07-09
fix volumiologrotate if subfolders exist inside LOG_FOLDER.

thegass:
after installation of squeezeboxserver i got error in journalctl because your logrotate doesn't recognize subfolders like /var/log/squeezeboxserver in LOG_FOLDER (/var/log/).
no idea if find is the best solution to only check files but it solves the error.


volumio:
Thanks!

PR #421, opened 2020-09-18
Build x86 tweak1

aldot:
a few tweaks and fixes, please apply!

PS: Should i be using something other than master? That one seems to use debian "jessie" (!) TIA


volumio:
Thanks!

aldot:
thanks.

watch out for the "debian-archive-keyring was listed twice" in the x86 config iff somebody copied the old jessie incarnation of the package list though ;)

cheers,

Issue #399, opened 2020-01-23
swapon fails with "Invalid argument"


kionez:
While I'm debugging some strange issues in my PiZero installation (mpd fails to start) I've noticed a error in system's log: when detects less than 512MiB of RAM it creates a swapfile of 512MiB in /data/swapfile and tries to enable with "/sbin/swapon /data/swapfile" but it fails because /data is mounted on a overlayfs which doesn't support reading pages so the -EINVAL error and "Invalid argument" returned.

`swapon: /data/swapfile: swapon failed: Invalid argument`

Edit: the file involved is: volumio/bin/dynswap.sh

volumio:
Interesting. This must be the result of something that changed on new kernels.
Then we need to mount the swap directly on the mountpoint and not on the overlayfs.
What do you think?

kionez:
Yes, creating the swapfile elsewhere on a filesystem which supports it should fix it, but I don't know where.. I don't know very well how the filesystems are organized. I think /media/volumio_data/ should be a good place (ext4), but seems created only in SD > 16GiB (I know.. I have to study the install scripts before talking :) ).

kionez:
> [cut] but seems created only in SD > 16GiB 

I dont know if I was wrong, I can't find anything in build scripts... I have a RaspberryPI3B+ with a 32GiB sdcard with mmcblk0p3 partition formatted and mounted and a RaspberryPIZero with a 16GiB sdcard without p3 partition (I've just flashed the last image and 2.692 image)... Maybe I have to dig better into code and avoid to go off-topic :)

volumio:
@gkkpch what do you think? Can we mount the swap directly on the partition without mounting it to overlayfs?

gkkpch:
I see no reason why not.
But, as this should be generic for all boards, we cannot assume that the data partition is on /dev/mmcblk0p3. 
dynswap.sh should determine which partition /data is mounted on, create a new mount point on /, e.g. /swap , mount the partition there and create the swapfile.

kionez:
Another idea: creating a dedicated swap partition should be faster and safer than using a swapfile but maybe it will works only on new installation because re-partitioning a running system could be a problem. 

(Anyway, now I'm trying to find why mmcblk0p3 is created but not mounted on my PiZero)

macmpi:
Just bumping into this as per [this thread](https://community.volumio.org/t/pi-zero-w-unresponsive/40847/22).

[...]

PR #426, opened 2020-11-14
fix dynswap.sh

macmpi:
as per issue https://github.com/volumio/Build/issues/399
`/data` is part of overlay, and can not be used for swap.
`swapon: /data/swapfile: swapon failed: Invalid argument`

Mount partition into new `/swap`, and enable swapfile as before


volumio:
very nice one. were you able to test it?

macmpi:
Yes, it works on my PiZero.
If you do not have such device to test on, you could change the memory size test number (512844) to sth exceeding your actual device memory size, and see swap being created.

volumio:
Thanks!

PR #428, opened 2020-12-06
Fixed Version typo in build.sh

rexjohannes:
Volumio3UI to Volumio2UI


volumio:
Actually, its not a typo. But I agree we shall report "Classic" and "Contemporary" UI to have better coherence

PR #430, opened 2021-01-06
Make supporting board families easier

ashthespy:
Handle board/device families easier. 
Until the platform-xxx repos have a uniform naming scheme, we need to provide more flexibility..
 
Handle cases like the Khadas [`kvim1|2|3` devices](https://github.com/volumio/Build/blob/168ba7444df98cd868db72f24faec0b6346489cc/build.sh#L375-L377) that all share a common [`vims` kernel/uboot tarball.](https://github.com/volumio/Build/blob/168ba7444df98cd868db72f24faec0b6346489cc/scripts/kvimsimage.sh#L82-L85)
```bash
# File vim3
### Device information
DEVICENAME="Khadas Vim3"
# This is useful for multiple devices sharing the same/similar kernel
DEVICEFAMILY="khadas"
# tarball from DEVICEFAMILY repo to use
DEVICEBASE="vims" # Defaults to ${DEVICE} if unset
DEVICEREPO="https://github.com/volumio/platform-${DEVICEFAMILY}"
```
Also allow for other cases where the [tarball](https://github.com/volumio/Build/blob/168ba7444df98cd868db72f24faec0b6346489cc/scripts/nanopineo2image.sh#L87) has a different name than the [device](https://github.com/volumio/Build/blob/master/build.sh#L3430. 
```bash
# File nanopineo2.sh
### Device information
DEVICENAME="NanoPi Neo2"
# This is useful for multiple devices sharing the same/similar kernel
DEVICEFAMILY="nanopi"
# tarball from DEVICEFAMILY repo to use
DEVICEBASE="nanopi-neo2" # Defaults to ${DEVICE} if unset
DEVICEREPO="https://github.com/volumio/platform-${DEVICEFAMILY}"
```


volumio:
Before merging, we shall check that the hw in os-release is consistent with current naming scheme.
@ashthespy @gkkpch can you please check?

ashthespy:
`VOLUMIO_HARDWARE` defaults to the name of the device as passed in using the `-d` flag. 
With the modular system, I just went from `<device>image.sh -> <device>.sh`
This avoids the [switch case in `build.sh`](https://github.com/volumio/Build/blob/master/build.sh#L237-L380), and instead directly uses `<device>.sh`
 
Alignment of names can be done using `VOL_DEVICE_ID`, when <device> doesn't correspond directly to the `VOLUMIO_HARDWARE`.

PR #435, opened 2021-01-10
Device check

xipmix:


gkkpch:
This fix is appears to be problematic when one only builds a rootfs (e.g. using "-b armv7").

ashthespy:
IIRC that was the reason for it being down there.. Else we need to check if -b was passed in this exit..

xipmix:
Doh! I'll take a look at how to improve the patch.

xipmix:
Hm I tried to build with -b armv8 and got loud complaints about no device being selected.
```
# ./build.sh -b armv8  -v buster

[ .... ] Running Volumio Image Builder -  
[ error ] No configuration found for <> 
[ .. ] Build system currently supports 10 devices: [ motivo mp1 nanopineo2 odroidn2 orangepilite raspberry rockpis tinkerboard x86_amd64 x86_i386 ] 
```

@gkkpch Can you please post your build attempt and the log showing the problem?

xipmix:
It may be enough to wrap that check in a check for $DEVICE being set
```
+if [[ -n "$DEVICE" ]]; then
   DEV_CONFIG="${SRC}/recipes/devices/${DEVICE}.sh"
   if [[ ! -f $DEV_CONFIG ]]; then
     log "No configuration found for <${DEVICE}>" "err"
     log "Build system currently supports ${#DEVICE_LIST[@]} devices:" "${DEVICE_LIST[*]}"
     exit 1
   fi
+fi
```

ashthespy:
Nope, when you do a `-b arm` to prebuild a rootfs, device is not set.. you need to check both :-)

volumio:
@xipmix @gkkpch @ashthespy 

What about reverting it? And then adding later on when tested?
I agree that while working on buster, let's wait to merge before carefully tested ;)

xipmix:
sure, revert & I'll respin

ashthespy:
@xipmix The idea behind this was to avoid a base `-b <>` build  if the supplied device `-d <>`  is not supported correct?

xipmix:
Not really - my idea was to stop the build entirely because nonsensical arguments were supplied.
I now understand that if it builds an image and then falls over due to an invalid device I can retry with ```-d``` and that just-image, bit it seemed better to just abort early.

Ge has taken another swing at this but it's not quite right, see 7fc0cb0a34a594c32e5c633f756b6da934de4b0c

gkkpch:
I think @xipmix idea is ok, especially considering the build server. There we always (re)build the rootfs and device in one go. So a typo means 15-20 minutes lost.

ashthespy:
Something like this -- I haven't had coffee yet though... ;-)
```diff
diff --git a/build.sh b/build.sh
index 7fd2b9b..74658db 100755
--- a/build.sh
+++ b/build.sh
@@ -217,6 +217,15 @@ function patch_multistrap_conf() {
   esac
 }
 
+function check_supported_device() {
+  DEV_CONFIG="${SRC}/recipes/devices/${DEVICE}.sh"
+  if [[ ! -f $DEV_CONFIG ]]; then
+    log "No configuration found for <${DEVICE}>" "err"
+    log "Build system currently supports ${#DEVICE_LIST[@]} devices:" "${DEVICE_LIST[*]}"
+    exit 1
+  fi
+}
+
 #Check the number of arguments. If none are passed, print help and exit.
 NUMARGS=$#
 if [ "$NUMARGS" -eq 0 ]; then
@@ -249,19 +258,13 @@ while getopts b:v:d:p:t:h: FLAG; do
     ;;
   esac
 done
-
-if [ ${DEVICE:=none} != none ]; then
-  DEV_CONFIG="${SRC}/recipes/devices/${DEVICE}.sh"
-  if [[ ! -f $DEV_CONFIG ]]; then
-    log "No configuration found for <${DEVICE}>" "err"
-    log "Build system currently supports ${#DEVICE_LIST[@]} devices:" "${DEVICE_LIST[*]}"
-    exit 1
-  fi
-fi
-
 # move past our parsed args
 shift $((OPTIND - 1))
 
+# Check if our device is supported if no Build flag was passed in
+[[ -z ${BUILD} ]] && check_supported_device  # No build flag
+[[ -n ${DEVICE} ]] && check_supported_device # Device flag (w/wo build)
+
```

PR #439, opened 2021-01-11
added PasswordAuthentication in sshd_config

soko1:
very important parameter for safety. let it be set to "yes" by default, and if desired, it can be set to "no" by adding your key


volumio:
Thanks for your contributions.
We are actively working in this branch: would you mind pushing your PRs there?
https://github.com/volumio/Build/tree/volumioOS/scripts

Then: let me clarify that wpasupplicant and the other files that were given 777 are required to be writable by user volumio. Please propose impros which satisfy such requirement. 
Thnks

soko1:
okay, no problem!

PR #442, opened 2021-01-11
fix permissions

soko1:
fix perm for /etc/wpa_supplicant/wpa_supplicant.conf, /etc/mpd.conf

### DIFF
--- a/scripts/volumio/configure.sh
+++ b/scripts/volumio/configure.sh
@@ -71,7 +71,8 @@ cp "${SRC}/volumio/etc/ssh/sshd_config" "${ROOTFS}/etc/ssh/sshd_config"
 
 #Mpd
 cp "${SRC}/volumio/etc/mpd.conf" "${ROOTFS}/etc/mpd.conf"
-chmod 777 "${ROOTFS}/etc/mpd.conf"
+chown mpd:audio "${ROOTFS}/etc/mpd.conf"
+chmod 660 "${ROOTFS}/etc/mpd.conf"
 
 #Log via JournalD in RAM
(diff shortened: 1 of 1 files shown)


volumio:
I say it again: 
let me clarify that wpasupplicant and the other files that were given 777 are required to be writable by user volumio. 
Please propose impros which satisfy such a requirement.

soko1:
I foresaw it! the user volumio is in the adm group, and the rights are set to it

soko1:
660 means that the user (root) and the group (adm) are allowed to write and read, for the rest (0) you can neither edit nor read the file

soko1:
```root@audiophile:/etc# chown root:adm /etc/wpa_supplicant/wpa_supplicant.conf
root@audiophile:/etc# chmod 660 /etc/wpa_supplicant/wpa_supplicant.conf
root@audiophile:/etc# su volumio
volumio@audiophile:/etc$ groups
volumio adm lp dialout cdrom floppy audio dip video plugdev netdev i2c input systemd-journal spi gpio
volumio@audiophile:/etc$ echo test >>/etc/wpa_supplicant/wpa_supplicant.conf
volumio@audiophile:/etc$ cat /etc/wpa_supplicant/wpa_supplicant.conf
ctrl_interface=/var/run/wpa_supplicant
network={
scan_ssid=1
ssid="dlink314"
psk="kocikacalujuurocikmikrotik"
priority=1
}
test```

:)

ashthespy:
`660` won't work, as our node process won't be able to write to it. The node process is run as `volumio:volumio`.
So `640` should work, but all this quite pointless if you ask me, as anyway everything is run using default username password that any "attacker" would obtain with a simple web search..

PR #443, opened 2021-01-11
fix permissions two

soko1:
I apologize, there was no way to check, but everything should work as it should and everything should change under the user of volumio. He is in the adm/audio group.
If there are mistakes, please let me know and I will correct.


ashthespy:
Could we hold on merging this until tested?

volumio:
@ashthespy sure. Let's add to the todolist in buster and I'll check it personally. once good to go we will merge it

ashthespy:
See https://github.com/volumio/Build/pull/442#issuecomment-774675128 this may not work as intended..

PR #447, opened 2021-01-13
Allow for device specific extra options for file system creation.

ashthespy:
This is useful to exclude ext4 metadata checksums on older kernels..
@gkkpch What do you think? Too much flexibility?


gkkpch:
Let's do this until we decide what to do with old kernel support.
Though Odroid C1 (kernel 3.10.107) now works, C2 suffers issues with systemd.journal. 
Not sure how much time I want to invest, but dropping it for the time being. 
If the community runs riot, we can always reconsider and add it later.

PR #449, opened 2021-01-13
Revert to raspberry -> pi device name, move uart flags to userconfig

ashthespy:
Fix kernel panics on non Debug images on pi3..

PS: test with 
```
git fetch origin pull/449/head:raspberry
git checkout raspberry
build.sh []
```


volumio:
@ashthespy still we're not there... the name shall be = pi .
This way folder structure is consistent...

ashthespy:
>the name shall be = pi .

How would you prefer this 
Set as before 
 -- `VOLUMIO_HARDWARE=pi` 
 or 
-- rename `raspberry.sh` -> `pi.sh`?

volumio:
-- rename raspberry.sh -> pi.sh?

thanks!

PR #450, opened 2021-01-13
Ramisk build WIP

volumio:


ashthespy:
I am not sure if this should be in the Build scripts - can't you do something similar directly with 
```bash
sudo mount -t tmpfs -o size=<> rootfs ./build
sudo mount -t tmpfs -o size=<> volmnt /mnt/volumio
./build.sh <>
sudo umount ./build
sudo umount ./mnt/volumio
```

PR #453, opened 2021-01-14
Streamline init-updater 

ashthespy:
Fetches from new repo location, warns when not found and proceeds with the build. 
@gkkpch  The actual updater binary (armv7) still needs testing :-)


ashthespy:
Will open a new PR with just this commit. Sorry!

PR #458, opened 2021-01-15
Use deb.debian.org instead of httpredir.d.o

xipmix:
https://wiki.debian.org/DebianGeoMirror states that httpredir has been discontinued.

I've tested this only by building a volumio image; not sure what further testing is required.

### DIFF
--- a/scripts/config.sh
+++ b/scripts/config.sh
@@ -12,7 +12,7 @@ declare -A SecureApt=(
 
 # Repo locations that are utilised to create source.list in the rootfs
 declare -A APTSOURCE=(
-  [Debian]="http://httpredir.debian.org/debian"
+  [Debian]="http://deb.debian.org/debian"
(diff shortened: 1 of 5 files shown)


ashthespy:
Ah -- while you are at it, can you remove https://github.com/volumio/Build/tree/volumioOS/volumio/etc/apt? 
Sources are created directly in [`configure.sh`](https://github.com/volumio/Build/blob/volumioOS/scripts/volumio/configure.sh#L16-L24) with [`config.sh`](https://github.com/volumio/Build/blob/volumioOS/scripts/config.sh#L13-L17) that you already updated..

xipmix:
not following. you mean ```git rm volumio/etc/apt/*``` ??

ashthespy:
Yes, `git rm -r volumio/etc/apt`

ashthespy:
Thanks! :-)

PR #463, opened 2021-01-15
Streamline mkinitramfs and volumio-init-updater

ashthespy:
This should streamline initrmafs creating a bit.

-- Add volumio-init-updater to all devices
-- Hopefully reduce initrd size a little
-- Speed up initramfs creation


volumio:
This shall be carefully tested before merging...

ashthespy:
I know the diff looks large -- but it's due to just a file rename (`mkinitramfs-buster -> mkinitramfs-custom`)
We have been using the same `mkinitramfs-buster` script since the start :-) 
Here is the changeset when comparing with the old `mkinitramfs-buster` https://www.diffchecker.com/AW1nLjbM

The main diffs are 
-- setting gzip level to -9 
-- streamline volumio-init-updater addition

volumio:
What is the advantage of -- setting gzip level to -9 ?

gkkpch:
I believe the most cases the best way to reduce mkinitramfs is with using modules= list instead of modules=most. X86 may be an exception, as we currently need to load many storage device driver modules. Perhaps we should have all storage & graphics modules built-in and then also use modules=list (which drops the rest), never tried that...

volumio:
We did have modules=list in the past, didn't we? 
what would be your ideal approach?

ashthespy:
It should reduce the initrd size (in theory), with a penalty w.t.r time to compress
[`man gzip`](https://linux.die.net/man/1/gzip)
>-# --fast --best
    Regulate the speed of compression using the specified digit #, where -1 or --fast indicates the fastest compression method (less compression) and -9 or --best indicates the slowest compression method (best compression). The default compression level is -6 (that is, biased towards high compression at expense of speed).

ashthespy:
> 
> 
> I believe the most cases the best way to reduce mkinitramfs is with using modules= list instead of modules=most. X86 may be an exception, as we currently need to load many storage device driver modules. Perhaps we should have all storage & graphics modules built-in and then also use modules=list (which drops the rest), never tried that...

It would great if we can get your modular initrd approach working on all devices! Probably once buster is settled?

Issue #429, opened 2020-12-15
Cleaning up redundant packages/config bits


ashthespy:
Using the Buster opportunity for some spring cleaning in the stuff copied into `rootfs` from [`/volumio`](https://github.com/volumio/Build/tree/master/volumio) folder in the repo.

In #413 I only looked into the `<device>config.sh` and `<device>image.sh` bits, and kept the [`volumioconfig.sh`](https://github.com/volumio/Build/blob/master/scripts/volumioconfig.sh) and [`configure.sh`](https://github.com/volumio/Build/blob/master/scripts/configure.sh) parts. 

From a quick exploration, here are some things to look at
### Services 
These are just [copied into the rootfs](https://github.com/volumio/Build/blob/13c723d98523acfdb1d282b7aa95a6b74d33cad8/scripts/configure.sh#L60-L61), but never used
- [ ] [`volumiosplash.service`](https://github.com/volumio/Build/blob/master/volumio/lib/systemd/system/volumiosplash.service)
- [ ] [`xbindkeysrc.service`](https://github.com/volumio/Build/blob/master/volumio/lib/systemd/system/xbindkeysrc.service)

### Irrelevant/Redundant configuration (?)
Not sure if these are still required?
- [ ] [`/etc/hostapd/hostapd.conf`](https://github.com/volumio/Build/blob/master/volumio/etc/hostapd/hostapd.conf) (not copied, and we append anyway in [`volumioconfig.sh`](https://github.com/volumio/Build/blob/13c723d98523acfdb1d282b7aa95a6b74d33cad8/scripts/volumioconfig.sh#L549-L564) to probably the default config?)
- [ ] [`/etc/inittab`](https://github.com/volumio/Build/blob/master/volumio/etc/inittab)
- [ ] [`/etc/netplug/*`](https://github.com/volumio/Build/tree/master/volumio/etc/netplug)
- [ ] [`/etc/hosts`](https://github.com/volumio/Build/blob/master/volumio/etc/hosts)
- [ ] [`/etc/nsswitch.conf`](https://github.com/volumio/Build/blob/master/volumio/etc/nsswitch.conf) Why is the default configuration overwritten? Do we need this?

Stuff that is handled by Node back end?
[...]

PR #466, opened 2021-01-24
Cleanup rootfs

ashthespy:
### Changes
-- Removing a lot of redundant multistrap config bits (like libraries, dummy packages) 
since we don't set [`omitrequired=true`](https://manpages.debian.org/testing/multistrap/multistrap.1.en.html#Restricting_package_selection), we can remove the libraries that are linked to installed packages.
>If the OmitRequired option is set to true, these packages will not be added - whilst useful, this option can easily lead to a useless rootfs. Only the packages specified manually in the configuration files will be used in the calculations - dependencies of those packages will be added but no others.

This then leaves extra libs that are really needed.  e.g `libavahi-compat-libdnssd-dev` for the [`mdns`](https://www.npmjs.com/package/mdns#installation) node package utilised by the back end. Ideally would like to move these all to a new section, to make tracking dependencies simpler. 
Currently we don't really know/recall what is required for what..

-- Removed some packages 
--- [`avahi-discover`](https://packages.debian.org/buster/avahi-discover) as we don't need the graphical browser.
--- [`udevil`](https://packages.debian.org/buster/udevil) as USB auto-mount is done using [`node-udev`](https://www.npmjs.com/package/udev)
---[`pyhton-requests`](https://packages.debian.org/buster/python-requests) not really sure what it was being used for?

There are quite a few more, like [`sqlite3`](https://packages.debian.org/buster/sqlite3) that [seems to be removed in the back-end](https://github.com/volumio/Volumio2/commit/2223fc5bbb64b8ad0a6348b3ca8c3c036fa01385) but still remain here, but I guess it's better to discuss/track these in #429 

--Streamline and log sizes of the rootfs so we see how it grows/shrinks.

N.B: All our custom `dpkg.cfg.d/` rules are not in place when we run multistrap, and I tried playing around getting multistrap to honour them, but didn't succeed. So all the `--path-exclude` stuff is only valid for installs that are done in chroot later on.. It would be great to not extract everything first, and then delete it in `finalize.sh`. Minor optimisation, but should think about it..

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -310,7 +310,7 @@ if [ -n "${BUILD}" ]; then
   log "Creating ${BUILD} rootfs" "info"
   #TODO Check naming conventions!
   BASE="Debian"
-  if [[ ! -f "$SUITE" ]]; then
+  if [[ -z "$SUITE" ]]; then
     log "Defaulting to release" "" "Buster"
     SUITE="buster"
   fi
--- a/recipes/base/VolumioBase.conf
+++ b/recipes/base/VolumioBase.conf
@@ -16,7 +16,7 @@ aptpreferences=/etc/apt/apt.conf.d/01progress
 
 #TODO: Prune and organise these packages!
 [Base]
-packages=acl apt apt-transport-https base-files base-passwd bash busybox debconf debconf-i18n debian-archive-keyring debianutils dmsetup e2fslibs e2fsprogs gcc-8-base gnupg gpgv initscripts insserv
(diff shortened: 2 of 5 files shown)


volumio:
@ashthespy please wait next time ;) I was still testing it...

ashthespy:
Oops sorry -- everyone seems to be committing directly to the branch, so wanted to get it in as well before stuff diverges..

volumio:
@ashthespy all looks good! well done!

PR #470, opened 2021-01-26
Remove redundant configuration bits

ashthespy:
Following up from #429 

- [x] `iptables.rules` isn't being used, and is being set directly by [`iptables.service`](https://github.com/volumio/Build/blob/volumioOS/volumio/lib/systemd/system/iptables.service)
- [x]  `xbindkeysrc.service` is redundant with the move to `triggerhappy` https://github.com/volumio/Build/pull/456
- [x] Remove `mpd.conf` in favour of  [`mpd.conf.tmpl`](https://github.com/volumio/Volumio2/blob/master/app/plugins/music_service/mpd/mpd.conf.tmpl)
- [x] Remove `./volumio/etc/fstab.x86`, we create it on-the-fly in [`chrootconfig.sh`](https://github.com/volumio/Build/blob/volumioOS/scripts/volumio/chrootconfig.sh#L28-L38)
- [x] Remove `./volumio/etc/ata-modules.x86 `, in favour of list in [`x86.sh`](https://github.com/volumio/Build/blob/63c09fb306755d018a9b9ec8d301264746dbf0a3/recipes/devices/families/x86.sh#L45-L53)
- [x] Remove `./volumio/etc/default/grub` as it's now fetched from [`platform-x86`](https://github.com/volumio/platform-x86/blob/master/packages-buster/efi/BOOT/grub.cfg) repo.
- [x] Remove `netplug` relics -- `volumio-network.service` and `volumio-network.d/netplug.conf`


ashthespy:
Round 1 should be ready to test and merge, will open a new PR with more clean-up soon :-)

PR #472, opened 2021-02-05
Create userconfig.txt on firstart only on Raspberry devices

ashthespy:
Few questions though, 

-- Does this script run on every upgrade? Or only on the first time the image is flashed and run? 
-- Why are we appending, shouldn't it rather check if the file exists and if not create it?

### DIFF
--- a/volumio/bin/firststart.sh
+++ b/volumio/bin/firststart.sh
@@ -1,15 +1,21 @@
-#!/bin/bash
+#!/usr/bin/env bash
+set -eo pipefail
+
+#shellcheck source=/dev/null
+source /etc/os-release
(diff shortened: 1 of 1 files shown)


gkkpch:
When firststart.sh runs after each update, than the use of it for userconfig.txt is mute.
Checking and not creating it would be a better approach in that case.

Creating the file was done in the old version of firststart.sh, but perhaps this is not working as well as I thought. Should be a test case for OTA.

ashthespy:
> -- Does this script run on every upgrade? Or only on the first time the image is flashed and run?

Did someone manage to test this out yet?

PR #474, opened 2021-02-07
Cleanup more redudndant `mpd.conf`

ashthespy:
Didn't push everything in #470.
Strangely the `mpd.conf` file was being copied twice for some reason, once in `build.sh` that was removed in 20e3c8c
and then again in `configure.sh` that this cleans up..


volumio:
Yep, saw it and fixed it (butdid not push )

From now on, let's try to push only tested PRs which are verified to work ;)

PR #475, opened 2021-02-09
X86: fix kernel image name to 'vmlinux'

gkkpch:
I suggest this one to replace #473

### DIFF
--- a/recipes/devices/families/x86.sh
+++ b/recipes/devices/families/x86.sh
@@ -158,16 +158,18 @@ device_chroot_tweaks_pre() {
   # Exact kernel version not known
   # Not brilliant, but safe enough as platform repo *should* have only a single kernel package
   # Confirm anyway
-  dpkg -i linux-image-*_"${ARCH}".deb
+  dpkg-deb -R linux-image-*_"${ARCH}".deb ./
+  rm -r /DEBIAN
+  rm -r /etc/kernel
+  rm -r /usr/share/doc/linux-image*
   rm linux-image-*_"${ARCH}".deb
-  log "Chanege linux kernel image name to 'vmlinuz'"
 
+  log "Change linux kernel image name to 'vmlinuz'"
   # Rename linux kernel to a fixed name, like we do for any other platform.
   # We only have one and we should not start multiple versions.
   # - our OTA update can't currently handle that and it blows up size of /boot and /lib.
-
   # This rename is safe, because we have only one vmlinuz in /boot
-  mv $(ls /boot/vmlinuz*) /boot/vmlinuz
+  mv /boot/vmlinuz* /boot/vmlinuz
 
   log "Preparing BIOS" "info"
   log "Installing Syslinux Legacy BIOS at ${BOOT_PART-?BOOT_PART is not known}"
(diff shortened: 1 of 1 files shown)


recipes/devices/families/x86.sh
```diff
@@ -160,17 +160,14 @@ device_chroot_tweaks_pre() {
...
-
-  ls -l /boot
+  log "Chanege linux kernel image name to 'vmlinuz'"
+
+  # Rename linux kernel to a fixed name, like we do for any other platform.
+  # We only have one and we should not start multiple versions.
+  # - our OTA update can't currently handle that and it blows up size of /boot and /lib.
+
+  # This rename is safe, because we have only one vmlinuz in /boot
+  mv $(ls /boot/vmlinuz*) /boot/vmlinuz
```
ashthespy, line 170: Shouldn't this just be `mv /boot/vmlinuz* /boot/vmlinuz`?

ashthespy:
Side note, it would be nice to add an explicit log line with the actual version of the kernel installed. Makes it easier to later check logs..

recipes/devices/families/x86.sh
```diff
@@ -246,7 +243,7 @@ EOF
   sed -i "s/%%IMGPART%%/${UUID_IMG}/g" ${grub_cfg}
   sed -i "s/%%BOOTPART%%/${UUID_BOOT}/g" ${grub_cfg}
   sed -i "s/%%DATAPART%%/${UUID_DATA}/g" ${grub_cfg}
-  sed -i "s/%%KRNL%%/${KRNL}/g" ${grub_cfg}
+
```
ashthespy, line 246: Wouldn't this lead to a nonbootable image -- [`grub.cfg`](https://github.com/volumio/platform-x86/blob/0e4dd9c56ad0c6224d7854685abdf03f454b2278/packages-buster/efi/BOOT/grub.cfg#L117) will try and load `%%KRNL%%` as the kernel.

gkkpch:
It already does that with the dpkg install and also shows the version number when building the initramfs.

recipes/devices/families/x86.sh
gkkpch: even simpler when that works, never thought of that

recipes/devices/families/x86.sh
ashthespy: Ah, https://github.com/volumio/platform-x86/commit/e9aad47c415c73b772e8542d63f02e1720130bae covers that! 👍

gkkpch: > Wouldn't this lead to a nonbootable image -- [`grub.cfg`](https://github.com/volumio/platform-x86/blob/0e4dd9c56ad0c6224d7854685abdf03f454b2278/packages-buster/efi/BOOT/grub.cfg#L117) will try and load `%%KRNL%%` as the kernel.

Yes, but the platform files were updated, this works fine

gkkpch:
Do we agree to merge this one?
(I know, the comments have a typo, I'll fix that later)

volumio:
@gkkpch if you confirm that all is ok and tested, I'll merge

volumio:
@gkkpch please confirm when I can merge (after its been tested)

gkkpch:
Yes, this was tested before pushing.

volumio:
Thanks!

PR #483, opened 2021-02-11
Tweak x86 partitions sizes/ Activate serial console for RPi while in buster test phase

gkkpch:


gkkpch:
Serial console on RPi is activated by build variable DEBUG_IMAGE=yes.
Tested!

gkkpch:
Please merge when this is OK

ashthespy:
> Serial console on RPi is activated by build variable DEBUG_IMAGE=yes.

Please be aware that this also switches the onboard Bluetooth module to use the mini UART. This might cause hard to trace issues with Bluetooth, so switch it back off before debugging any Bluetooth stuff.

volumio:
Thanks!
@ashthespy is there a way to have uart enabled without impacting BT?
@gkkpch for next times, one PR for one thing (so it could have been splitted in 2 prs ;) )

gkkpch:
agree, that was not intentional...

ashthespy:
>  @ashthespy is there a way to have uart enabled without impacting BT?

Not too sure, reading https://www.raspberrypi.org/documentation/configuration/uart.md suggests you could enable `uart2` with an overlay and switch the console to that. Not sure if it is worth the effort..

gkkpch:
if it is too complicated then perhaps have two build options, one for serial only and the other one for bt?

ashthespy:
> 
> 
> if it is too complicated then perhaps have two build options, one for serial only and the other one for bt?

Hmm, wouldn't that just be `DEBUG_IMAGE` -- because all that does it enable the UART? 
I might be confusing issues here - but how I understood it was that for console to be on UART, you need to move Bluetooth to the `miniuart` and they both can't be on the same one..

gkkpch:
is that the reason? then it is not a solution

Issue #479, opened 2021-02-10
Systemd and volumio files not properly updated


volumio:
@ashthespy 

Issue:
- The custom files in https://github.com/volumio/Build/tree/volumioOS/volumio are copied over too soon.
- When installing custom packages https://github.com/volumio/Build/blob/volumioOS/build.sh#L548 for example, the systemd file in the custom package overwrites the volumio custom one

GOAL:
- Have the files  https://github.com/volumio/Build/tree/volumioOS/volumio copied as late as possible in the build process, following the pattern used in https://github.com/volumio/Build/blob/master/scripts/configure.sh
- This will make sure that all the customization we've done on those files are respected
- Please send a PR (tested) to volumioOS

ashthespy:
This is something I wanted to clarify -- this part is from the old build system that I copied over, but it would be good to refactor.


#### Rootfs Creation
-- Run `multistrap` ([volumioOS](https://github.com/volumio/Build/blob/7e70a9a70ccd64d4428d7e4fd88b97d5d2d1a8ba/build.sh#L367)  [master](https://github.com/volumio/Build/blob/e81b2037a65d8f554cb0c857a1007d659ad7f838/build.sh#L162))
-- Run `volumioconfig.sh` in chroot ([volumioOS](https://github.com/volumio/Build/blob/7e70a9a70ccd64d4428d7e4fd88b97d5d2d1a8ba/build.sh#L395) [master](https://github.com/volumio/Build/blob/master/build.sh#L201))
-- Run `configure.sh` on the resulting rootfs ([volumioOS](https://github.com/volumio/Build/blob/7e70a9a70ccd64d4428d7e4fd88b97d5d2d1a8ba/build.sh#L419) [master](https://github.com/volumio/Build/blob/master/build.sh#L227))


#### Image creation
 -- Run [`<device>image.sh`](https://github.com/volumio/Build/blob/e81b2037a65d8f554cb0c857a1007d659ad7f838/build.sh#L240) ([`makeimage.sh`](https://github.com/volumio/Build/blob/7e70a9a70ccd64d4428d7e4fd88b97d5d2d1a8ba/build.sh#L561) in volumioOS
[...]

PR #484, opened 2021-02-11
Workaround for configuration files for  custom packages not installed…

ashthespy:
… during rootfs creation

Fixes https://github.com/volumio/Build/issues/479

PS: Can't test ATM but should work in theory.


volumio:
That's not going to fix the issue... We have the same issue with shairport...

ashthespy:
Hmm, currently we use upstream shairport, so it's pulled in via multistrap. When you add the custom version, you could  copy it's service file here similarly as well?  If you blanket copy all the service files you would have to test to ensure other issues don't crop up.  If you ask me, this is the least effort solution to fast track things. I unfortunately currently don't have the time now to refactor things as discussed in #479

volumio:
@ashthespy ok let's implement this workaround since you don't have time, good suggestion ;)

PR #486, opened 2021-02-11
Odroid N2: fix uInitrd arch

gkkpch:
still wip


gkkpch:
We can skip this one.
With #487, which is preferred,  this would be obsolete anyway.

PR #488, opened 2021-02-11
Add kill and i2cset to the sudoers NOPASSWD list

timothyjward:


scripts/volumio/volumioconfig.sh
```diff
@@ -169,7 +169,7 @@ log 'Adding Safe Sudoers NoPassw permissions'
...
+volumio ALL=(ALL) NOPASSWD: /sbin/poweroff,/sbin/shutdown,/sbin/reboot,/sbin/halt,/bin/systemctl,/usr/bin/apt-get,/usr/sbin/update-rc.d,/usr/bin/gpio,/bin/mount,/bin/umount,/sbin/iwconfig,/sbin/iwlist,/sbin/ifconfig,/usr/bin/killall,/bin/ip,/usr/sbin/service,/bin/journalctl,/bin/chmod,/sbin/ethtool,/usr/sbin/alsactl,/bin/tar,/usr/bin/dtoverlay,/sbin/dhclient,/usr/sbin/i2cdetect,/sbin/dhcpcd,/usr/bin/alsactl,/bin/mv,/sbin/iw,/bin/hostname,/sbin/modprobe,/sbin/iwgetid,/bin/ln,/usr/bin/unlink,/bin/dd,/usr/bin/dcfldd,/opt/vc/bin/vcgencmd,/opt/vc/bin/tvservice,/usr/bin/renice,/bin/rm,/bin/kill,/usr/sbin/i2cset
```
ashthespy, line 172: Btw, why is `/opt/vc/bin/tvservice` and `/opt/vc/bin/vcgencmd` being added to all devices? 
Do we actually even use it?

timothyjward: `/opt/vc/bin/tvservice` is useful for disabling HDMI, which can save quite a bit of power (25mA) on the device. Quite useful if you're running on battery power.

ashthespy: Sure, but isn't it Raspberry Pi specific? This file is used in all {x86,x64,arm,armv7 and armv8} rootfs creation..
Furthermore, it is not something that the Node backend utilises that it needs to be in the NOPASSWD list? 

Sorry, I know it is unrelated, but just saw it while looking at the diff and was wondering..

PR #490, opened 2021-02-12
Add OrangePi H3 based family

ashthespy:
Finalise support for OrangePiOne and OrangePiPC.
Bumped from the sunxi's 5.4 to 5.10.y kennel and my OpiLite still works fine. 
The Opi One and PC I don't have to test, but it looks like @Vyacheslav-S has already [provided some test Buster images](https://community.volumio.org/t/volumio-debian-buster-beta-orange-pi-images/44826) although I am unsure with which platform files.


ashthespy:
Switched to Volumio specific repo -- but will be on `4.14.34` kernel and needs https://github.com/volumio/platform-orangepi/pull/6 to be merged for `5.10.y` kernel..

ashthespy:
@gkkpch When you get some time could you merge https://github.com/volumio/platform-orangepi/pull/6 and then this in? 
I'm currently running things on an OpiLite, but can't test on the rest that I don't own. The images do build if people with the device(s) are keen to test -- but don't think we should add this to the ongoing betas. It is a community device after all :-)

ashthespy:
Merged this in, been tested by people over in the forums as well, so should be fine :-)

PR #491, opened 2021-02-12
Init.x86

gkkpch:
Raising the limit for the boot püartition so we can edit volumio.initrd and have some working space to keep a copies


volumio:
I think this was sent by mistake?

ashthespy:
You can just change it to volumioOS branch...

PR #493, opened 2021-02-14
[CI] Switch to action-shell@v1

ashthespy:
Just realised that the CI stuff I set up over on [`ashthespy/VolumioOS`](https://github.com/ashthespy/VolumioOS/runs/1694777538?check_suite_focus=true) was also here. I moved the action out into it's own [action-shell](https://github.com/ashthespy/action-shell) repo, to not clutter stuff here. 
It's been updated to run on commits and PRs to the `volumioOS` branch.


volumio:
Nice one! 
How do we set up that in Volumio as well? 
I can set up a repo in Volumio, what would be a good naming for it?
volumio-ci-action-shell ?

ashthespy:
> How do we set up that in Volumio as well?
> I can set up a repo in Volumio, what would be a good naming for it?
> volumio-ci-action-shell ?

I don't follow, it's already setup for to run on the `volumioOS` branch of this repo, nothing further has to be done.
I just pulled it out to reuse it for other repos. Since the [action is public](https://github.com/marketplace?type=actions), anyone can utilise it.

volumio:
@ashthespy sorry for the late reply... before merge IMHO let's have the action repo on volumio ;)
suggest a name, I'll create it and grant you push access

ashthespy:
Sorry if I am being daft, but I still don't get it, there is nothing Volumio specific to this? In fact we already use this in other internal projects..

volumio:
Move this https://github.com/ashthespy/action-shell to a volumio repo, so also others can contribute sending PRs to Volumio. if you agree...
The reason is we have to keep all code that runs volumio under volumio's repo, for visibility and coherence

ashthespy:
>  The reason is we have to keep all code that runs volumio under volumio's repo, for visibility and coherence

I guess this is why I am confused, this is not something that "runs volumio". It's not a step in the build scripts, nor a package/dependency for the final image  -- it is just a GitHub CI action to run some static analysis and lint shell scripts.. 

If you really want to maintain it a Volumio specific copy, then would recommend closing this PR and leaving the stuff I wrote from  [`ashthespy/VolumioOS`](https://github.com/ashthespy/VolumioOS) in this repo as well. You would also then need to fork the rest of the pipeline steps, and then update the pipeline to point to the Volumio specific repos, such as:  
[`actions/github-script`](https://github.com/actions/github-script)
[`actions/checkout`](https://github.com/actions/checkout)
You could also re-write things as [composite action](https://docs.github.com/en/actions/creating-actions/creating-a-composite-run-steps-action) if you further want to remove external dependencies of the Alpine Docker image to run `shellcheck`.

Again, at the end of the day I don't see the point of all this, keeping a Volumio copy also adds the burden of maintain that Volumio copy - but it's your call! :-)

ashthespy:
Closed this for now, @volumio feel free to reopen it if required down the line :-)

PR #498, opened 2021-02-27
Fix ucm default issue

gkkpch:
It seems we were running after the facts.
Our own ucm folder is no longer needed with buster as the installation of Alsa tools already appears to store the defaults in /usr/share/alsa/ucm. Our (old jessie) version was just overriding and disabling Intel Atom soc audio.


gkkpch:
Safe to merge, only affects Intel Atom configurations (still WIP, but step forward)

PR #501, opened 2021-03-01
Fix missing dcfldd for installToDisk

gkkpch:
This also needs [Volumio2 PR #2087](https://github.com/volumio/Volumio2/pull/2087)


volumio:
Watch out, this overwrites the custom mpd which Tim sent 

Please just add the missing binary

PR #502, opened 2021-03-03
Add missing dcfldd for installToDisk (legacy method using dd)

gkkpch:


gkkpch:
This is a correction to replace #501

PR #505, opened 2021-03-15
init.x86: only expand glob to valid device names, return null otherwise

gkkpch:
This fix prevents an unchanged glob to be used as a device name, which causes this error at startup with installations which boot from eMMC or NVMe installed with no usb stick or other usb devices attached:
```
 lsblk: /dev/sd[a-z]: not a block device
```

### DIFF
--- a/scripts/initramfs/init.x86
+++ b/scripts/initramfs/init.x86
@@ -9,6 +9,7 @@
 #
 #
 
+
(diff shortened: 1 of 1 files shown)


ashthespy:
Sorry I missed this -- see https://github.com/volumio/Build/commit/603b3e78bd5c7c6ab8cd4c4244dc4b35f8a36a29#commitcomment-49403815

PR #507, opened 2021-04-12
Include default mpd conf

volumio:


ashthespy:
Were you able to test the suggestion from https://github.com/volumio/Build/commit/20e3c8c821896d19a48b8e9681de072f39aceb18#commitcomment-48138201 ? 

It should be as  simple as adding into `finalize.sh`
```bash
systemctl disable mpd.service
````

With this, you are still going to have race conditions + dual configurations that are going to go out of sync..

**Comment by volumio** [comment]
We already tried, but did not work because:
- Mpd is started anyway when the socket is initialized, resulting in a deadlock
-  This deadlock generated also a corrupted db

**Comment by gkkpch** [comment]
This will not work, we already tried it. As soon as a usb drive is mounted, an "mpc update" is issued which starts mpd anyway (via socket??). So we keep being too late with the mpd.conf correction. I believe Mi fixed it

**Comment by ashthespy** [comment]
ah, that is a shame!  Thanks for the clarification. 👍

**Comment by ashthespy** [comment]
I was reading something else and came across [this](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=768208) - FWIW, the socket is a separate service `mpd.socket`, and probably also needs to be stopped.

**Comment by ashthespy** [comment]
After a MPD upgrade, I got frustrated with the double configuration templates, so looked into this a bit more. 

> Mpd is started anyway when the socket is initialized, resulting in a deadlock

No, MPD is started when a connection is initiated via the socket, not on socket initialisation. 
i.e if `mpd.service` is stopped, but `mpd.socket` is started, then any connection to `/run/mpd/socket` triggers `mpd.service`

>As soon as a usb drive is mounted, an "mpc update" is issued

I don't believe this is a feature of `mpc`? I couldn't find it anywhere. 
I would guess it's something the mount handler in the volumio-core does this right?


This is what I have that seems to work in my case
From the build side:
```bash
systemctl stop mpd.service mpd.socket
```

Then from volumio-core side
```diff
diff --git a/app/plugins/music_service/mpd/index.js b/app/plugins/music_service/mpd/index.js
index b6808a6b..e0ad0db9 100644
--- a/app/plugins/music_service/mpd/index.js
+++ b/app/plugins/music_service/mpd/index.js
@@ -616,6 +616,9 @@ ControllerMpd.prototype.mpdInit = function () {
         self.logger.error('Could not create MPD File on system start: ' + error);
         defer.resolve();
       } else {
+        // Start mpd's socket so it can be started by libMpd.connect's call or subsequent service calls
+        // But start it only if it's not already started. Some NodeShell to the rescue (should be posix and not bash)
+        execSync('[ -S /run/mpd/socket ] || /usr/bin/sudo /bin/systemctl start mpd.socket', { uid: 1000, gid: 1000, encoding: 'utf8' });
         self.restartMpd(function (error) {
           if (error !== null && error != undefined) {
             self.logger.error('Cannot start MPD on system Start: ' + error);
```

This essentially doesn't let MPD start (the first time) before the configuration file is written by volumio-core.
It also starts the `mpd.socket` service, for subsequent `libMpd` calls.. 

PS:  `mpd/index.js` is such a mess and can really do with some ❤️!

ashthespy:
> I would guess it's something the mount handler in the volumio-core does this right?

Yep, here you go [`networkfs/index.js`](https://github.com/volumio/volumio3-backend/blob/d53720430fb30d19e6c26dd6d94cee2296edc508/app/plugins/system_controller/networkfs/index.js#L297-L309) it should be straightforward to get that function to use [`updateDb`](https://github.com/volumio/volumio3-backend/blob/eaf4f3eb489390050112cfead7e5cb3148fd96b9/app/plugins/music_service/mpd/index.js#L1849) instead.

volumio:
This solution seems interesting, can you do a PR?
Remember to always wrap sync functions in try catch ;)

PR #509, opened 2021-04-12
Enable rt5640 jackdetect

gkkpch:
Enable jack detection for Intel Atom RT5640-based sound devices. 
It allows switching between speaker and headphones when plugging/unplugging the headphone jack


gkkpch:
Prerequisite: https://github.com/volumio/platform-x86/commit/2804a2305f14cd2c36968505efb98283915c8d3d

PR #512, opened 2021-07-11
Introduce kernel module for WaveShare WM8960 Audio HAT

hftsai256:
WaveShare didn't provide pre-built binaries for raspberry pi and therefore I took their source code and patch it up to build required kernel modules under `device_chroot_tweaks_post()`.

I wonder running full `volumio kernelsource` command is an overkill because I only need headers to build `dkms` modules. Also, this is a quick-and-dirty hack for demonstration purpose, so please advice if there's a better place for it.

My magic line to `/volumio/app/plugins/system_controller/i2s_dacs/dacs.json` is:
```
{"id":"wm8960-soundcard","name":"Waveshare - WM8960","overlay":"wm8960-soundcard,i2s-mmap","alsanum":"1","mixer":"","modules":"","script":"","i2c_address":"1a","needsreboot":"yes"},
```

Thanks,
Halley

### DIFF
--- a/recipes/devices/pi.sh
+++ b/recipes/devices/pi.sh
@@ -369,7 +369,21 @@ device_chroot_tweaks_pre() {
 # Will be run in chroot - Post initramfs
 device_chroot_tweaks_post() {
 	# log "Running device_chroot_tweaks_post" "ext"
-	:
+	log "Download Kernel Sources"
+	volumio kernelsource
+	log "Install WM8960-Audio-HAT Kernel Module & Overlay"
+	git clone https://github.com/waveshare/WM8960-Audio-HAT.git /opt/WM8960
+	pushd /opt/WM8960
+	# hack system environment
+	sed -i 's/is_Raspberry=.*/is_Raspberry=Raspberry/g' install.sh
+	sed -i 's/uname_r=.*/uname_r="5.4.83+"/g' install.sh
+	# revert API to kernel version before 5.10.y
+	sed -i 's/snd_soc_component_read/snd_soc_component_read32/g' wm8960.c
+	sed -i '/\.no_capture_mute/d' wm8960.c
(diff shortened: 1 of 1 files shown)


hftsai256:
After several tries this script can build modules, but is having trouble to install it properly. The HAT is not enabled on a freshly built image:
```
volumio@voluzero:~$ aplay -l
**** List of PLAYBACK Hardware Devices ****
card 0: b1 [bcm2835 HDMI 1], device 0: bcm2835 HDMI 1 [bcm2835 HDMI 1]
  Subdevices: 8/8
  Subdevice #0: subdevice #0
  Subdevice #1: subdevice #1
  Subdevice #2: subdevice #2
  Subdevice #3: subdevice #3
  Subdevice #4: subdevice #4
  Subdevice #5: subdevice #5
  Subdevice #6: subdevice #6
  Subdevice #7: subdevice #7
volumio@voluzero:~$ sudo dkms status
[sudo] password for volumio: 
wm8960-soundcard, 1.0: added
```
However, follow the flow and re-run the `install.sh` script again will stage those files properly. Is this a QEMU thing? I'm not quite familiar with it.

ashthespy:
Thanks for this PR @hftsai256  -- but normally we don't build kernel modules/DAC firmware in the image builder.

Ideally firmware is compiled once for each kernel version, and then just fetched by the image builder on each run..
See here how other repos are structured to get an idea:
https://github.com/volumio/Build/blob/bc89b476e25e385d3cf4c9155471e47f5934028e/recipes/devices/pi.sh#L164-L170

And then the firmware tarballs are just fetched and added to the image:
https://github.com/volumio/Build/blob/bc89b476e25e385d3cf4c9155471e47f5934028e/recipes/devices/pi.sh#L229-L238

PR #513, opened 2021-07-13
Fetching pre-built artifacts for wm8960 DAC

hftsai256:
This PR is re-created as discussed in the previously-closed one: [PR#512](https://github.com/volumio/Build/pull/512). The referred archive is manually collected for now, and we may need a CI/CD system for our best interest.


ashthespy:
>tar cannot change ownership to uid xxx, gid xxx

Just add `--owner=0 --group=0` when you package it :-)

ashthespy:
With https://github.com/raspberrypi/linux/issues/4454 now fixed, we should be able to use the upstream overlay right and modules from 5.10.50 onwards.

However, we are on the latest kernel 5.4.yy currently available via  `rpi-firmware` (5.4.83), and since they have moved to 5.10.yy, don't think we are going to get prebuilt 5.4.yy kernel with the module included. So we would still always need to build the `snd-soc-wm8960` module externally as long as we are on 5.4.yy. 
Which considering https://github.com/raspberrypi/linux/issues/4108, might be some time :-( 

So tl;dr; lets merge this, and remove it when we move to 5.10.yy.

PR #515, opened 2021-08-13
Modular init scripts

ashthespy:
Looking at the `init` scripts, there is a lot of scope of simplification and having a modular approach. I believe @gkkpch already has done some work for the Installer images towards this. 

Ideally I reckon we can just consolidate all these into `init.vol` with the current approach.
```
./scripts/initramfs/
├── init
├── init.nextarm
├── init.nextarm_tvbox
└── init.x86
# to just this
├── init.vol
└── vol
    └── inithelpers
```
If there truly things the need to be done at the board level, it should be possible to add things that are defined in `<device>.sh`  that end up getting called from the `init` script. 

Currently this is just a start, things are quite rough still -- I have reworked some bits
-- added debugging
-- module loading
-- Fstab `\boot` updating 
-- Data part resizing 

Still have to do the USB, update and MCU stuff, but I don't have devices to test them so..


gkkpch:
Correct, I have a fully redesigned initramfs covering x86, PI and all other arm boards. For x86 and some non-PI boards testing has been a success, but I got stuck, due to lack of time (and motivation).
Hope to finish this in Autumn, may need some optimisations (i'm not the best coder).
Our initramfs handling is much closer to the way the official Debian/ Ubuntu initramfs tools work.

gkkpch:
you'll find my work in branch vOS-initramfs

ashthespy:
Wow, that is really a comprehensive piece work! I will use that as a base, and merge these bits in there..

Btw, I used a similar `edit-initramfs.sh` based on your work on [`init-edit.sh`](https://github.com/volumio/Volumio2/blob/master/app/plugins/system_controller/volumio_command_line_client/commands/init-edit.sh). But seemed to be running into issues with the `cpio` step. 

```
Creating a new uInitrd, please wait...
cpio: File ./uInitrd.new grew, 13120452 new bytes not copied
```

Any ideas? Could it be because of the version of `cpio` being used isn't the same? (Ubuntu LTS 2.13 vs Buster 2.12)?

gkkpch:
about the cpio issue: not sure, looks like it is trying to re-use the same file. And when you delete the original before running cpio and then create new? Did cpio get new options in buster and does not expand as a default when writing anew?

gkkpch:
post edited....

ashthespy:
Thanks for the tips, got it working now.  -- [`init-edit.sh`](https://gist.github.com/ashthespy/5e856e832c3904bc600f22c61b0f8766)  was indeed recopying the old `uInitrd` into the new image. 

I am going to close this PR and work on your branch in the coming weeks :-) Enjoy your vacation!

